PR #17, opened 2013-12-19
Updating to use B-Cycle API

jasongdove:
Quick update to switch from scraping to API. Data appears to be updating correctly.

### DIFF
--- a/pybikes/bcycle.py
+++ b/pybikes/bcycle.py
@@ -2,16 +2,19 @@
 # Copyright (C) 2010-2012, eskerda <eskerda@gmail.com>
 # Distributed under the AGPL license, see LICENSE.txt
 
-import re
+import json
 
-from pyquery import PyQuery as pq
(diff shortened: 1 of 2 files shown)


bparmentier:
API url is at http://api.bcycle.com/services/mobile.svc/ListKiosks now.
What's the status on this PR? Do we continue scraping the HTML page?

eskerda:
As far as I remember (2013), the API license that must be signed to get an API key included some sections regarding Copyright and redistribution that are incompatible with the project.
* no transmission and broadcasting of BCycle's data is permitted: meaning, the BCycle API is meant only to develop clients.
* Unclear claims on providing BCycle with licensing and trademarks.

Maybe the licensing terms have changed.

eskerda:
For completeness, we could provide an alternative official API way of accessing the data on pybikes (if that was useful for people using pybikes) but for the purpose of the CityBikes service, the API is not eligible for use.

PR #31, opened 2015-01-07
Add Velo Antwerpen

bparmentier:
Hope this is enough.

However, I'm not sure about latitude and longitude. I took it from Wikipedia (Antwerpen), but is it the center of the city or the center of the bike network?


eskerda:
Both work, usually preferring the center of the city and only using the center of the bike network when it makes a lot of sense (residential bike network, etc).

:+1:

PR #32, opened 2015-01-08
Add Pronto Cycle Share

colinmcglynn:
This is my first pull request to this repository so please let me know if I missed something. Thanks!

### DIFF
--- a/pybikes/data/bixi.json
+++ b/pybikes/data/bixi.json
@@ -174,6 +174,22 @@
             "format": "json"
         }, 
         {
+          "tag": "pronto-cycle-share", 
+          "meta": {
+            "city": "Seattle, WA", 
+            "name": "Pronto Cycle Share", 
+            "country": "US", 
+            "company": [
+              "Pronto!", 
+              "Alta Bicycle Share, Inc"
+            ], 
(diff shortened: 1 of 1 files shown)


eskerda:
Unfortunately this feed uses a different syntax not implemented in any of the parsers present in [bixi.py](https://github.com/eskerda/PyBikes/blob/experimental/pybikes/bixi.py)

Only way to support it is by adding _yet another_ parser in:
* parse_methods
* BixiSystem
* BixiStation

Thanks for contributing and feel free to ping me if you need further help!

Note to self, if this becomes a common problem (more Alto systems implementing different schemes) we might refactor the code to make it easier / cleaner to add systems to bixi.

bparmentier:
If I'm not mistaken, the Pronto scheme is not operated by Bixi: http://www.seattlebikeblog.com/2014/04/03/seattle-based-system-will-pioneer-new-bike-share-partnership-hiring-manager/

Also check the conversation on this tweet for tags hints: https://twitter.com/CyclePronto/status/521773054280482816

eskerda:
Created issue about adding Pronto: https://github.com/eskerda/pybikes/issues/68

Issue #21, opened 2014-07-12
Real address


psyrax:
Addresses for some networks are wrong, for instance in ecobici 251 GOETHE-DARWIN, should be Goethe 17, Anzures, Miguel Hidalgo, 11590 Mexico City, Federal District, Mexico. I know that came from the original source but using reverse geocoding api from google it could be possible to catch the real addresses as in this example: http://ecobicidx-20724.onmodulus.net/stations/all

PR #33, opened 2015-01-10
Add Veloway system

bparmentier:
I'm not sure how to handle the encoding of the `wcom` tag. Could you give it a try?

### DIFF
--- a/pybikes/__init__.py
+++ b/pybikes/__init__.py
@@ -42,6 +42,7 @@ from .bicicard import *
 from .nextbike import *
 from .samba import *
 from .ciclosampa import *
+from .veloway import *
 
 __all__ = base.__all__ +\
           bixi.__all__ +\
@@ -58,7 +59,8 @@ __all__ = base.__all__ +\
           bicicard.__all__+\
           nextbike.__all__+\
           samba.__all__+\
-          ciclosampa.__all__
+          ciclosampa.__all__+\
+          veloway.__all__
 
 def getDataFiles():
     return resource_listdir(__name__, 'data')
--- a/pybikes/data/veloway.json
+++ b/pybikes/data/veloway.json
@@ -1,0 +1,50 @@
+{
+    "instances": [
+        {
+            "tag": "velobleu",
+            "meta": {
+                "latitude": 43.695949,
+                "city": "Nice",
+                "name": "Vélo Bleu",
+                "longitude": 7.271413,
+                "country": "FR"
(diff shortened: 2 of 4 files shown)


eskerda:
Seriously, indeed, WTF at ```wcom```, wouldn't touch that with a long poking stick.No wonder it takes a human life to render a map on their website. Thinking about adding a Hall of Shame of different gems we find on our journey. 

Note that some feeds do not include ```wcom``` (well, as null), failing at fix_name and not appending it to the stations list. For instance, vel'in just reports back 1 station
```
$ virtualenv2 v
$ source v/bin/activate
(v) $ sh runtests.sh TestSystems.test_veloway
[...]
Vélo Bleu has 166 stations
↑ stations look ok ↑                          

DK'Vélo has 31 stations
↑ stations look ok ↑                          

Vélocéa has 24 stations
↑ stations look ok ↑                          

Vél'in has 1 stations
↑ stations look ok ↑  
.
Ran 1 test in 4.778s

OK
```
* My recommendation would be to just drop  ```wcom```, as what it does provide is rather funky, varying from a human-readable address to just ```17 Station```. There's an open issue (#21) about adding a cohesive address field to all stations. 
* Should we drop stations that have neutral set to 1 (assuming that could happen)?
* I can't grasp the difference between ```ac``` and ```tc```.

bparmentier:
* Hey, got it working (I think) in https://github.com/bparmentier/PyBikes/commit/65b101ca354167d5e44fff48b5893cde3f9791d6. We could have dropped the `wcom` tag but the problem remains in `name` anyway…
* We could keep `neutral`, but I don't see any real difference with `disp`. Should be discussed.
* As I understand it, `tc` is the number of slots physically present, while `ac` is the number of working slots (some may be deactivated/broken/malfunctionning).

eskerda:
Good, thanks. Note that ``` if foo is not None and foo != ""``` is equivalent to ```if foo```

bparmentier:
> Note that if foo is not None and foo != "" is equivalent in python to if foo

Fixed.

Maybe should I also drop `extra['slots_available']` and just use `ac` for `extra'[slots']`. What do you think?

eskerda:
If the info is there, I would use it. On another feed we've got ```faulty_slots```, but I think ```slots_available``` makes sense in this case.

eskerda:
And off we go! 4 new bike sharing schemes. Many thanks!

PR #34, opened 2015-01-14
Add more cities to Bicincitta

bparmentier:
Add 28 cities to Bicincitta, 4 of which where moved from `bicincittaold`.

I also removed the trailing colon in `extra['description']`.


eskerda:
Awesome. This puts our network count to 220. You just broke the 200 record line! :trophy:

Linking to #28 for reference

PR #40, opened 2015-01-15
Add WE-cycle (Aspen) and CoGo (Columbus) to Bixi

bparmentier:
We also should replace *Alta Bicycle Share, Inc* with *Motivate* (http://www.motivateco.com/news/2015/01/14/alta-bicycle-share-becomes-motivate).


eskerda:
We can s/Alta Bicycle Share, Inc/Motivate

PR #46, opened 2015-03-24
German Call-A-Bike system

neffs:
Initial version for Call-A-Bike system in Germany.

### DIFF
--- a/pybikes/__init__.py
+++ b/pybikes/__init__.py
@@ -45,6 +45,7 @@ from .ciclosampa import *
 from .veloway import *
 from .easybike import *
 from .cleanap import *
+from .callabike import *
 
 __all__ = base.__all__ +\
           bixi.__all__ +\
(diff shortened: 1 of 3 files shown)


eskerda:
Hello, thanks for your contribution!

I will take a bit to merge it on the service, as to check how the API reacts to a system were stations have no ```slots``` set.

Do you have any input about how Call-A-Bike works and more call-a-bike networks to add? Are all stations parsed real stations or some might be bikes (flexbike)?

Thanks

neffs:
Slots are never relevant, but otherwise the call-a-bike systems are different in most cases. In some cities it can be only a single (flex)bike.

Some examples:
Berlin: Fixed stations with slots and a terminal, but bikes use a short range wireless system. If slots are full the bike can be parked next to the station.
Stuttgart: Like Berlin but with additional e-bikes which are only available and returnable at some stations.
Frankfurt: Bikes have GPS localization and a GSM modem. Fixed stations consist of a street sign and a GPS marker, no terminal or other infrastructure. In the inner city you can return your bike at every intersection for an additional fee.
Cologne: Bikes can be returned at every intersection in a fixed area. No fixed stations at all.

There are more cities which have a single call-a-bike station.

PR #50, opened 2015-05-11
Added KVB Rad Germany

madflow:
+ Added https://nextbike.net/maps/nextbike-live.xml?domains=kg


eskerda:
Thanks for contributing!

I've removed 'Germany' from the bike share name.

madflow:
Thanks!

PR #51, opened 2015-07-10
Fix bcycle

mstrcnvs:
Though we still have a failing test for the Hawaii B-cycle, which for some reason has no stations on their site. What can we do about it so bcycle passes all tests?

### DIFF
--- a/pybikes/bcycle.py
+++ b/pybikes/bcycle.py
@@ -6,6 +6,7 @@ import re
 
 from lxml import html
 from .base import BikeShareSystem, BikeShareStation
+from pybikes.exceptions import InvalidStationException
 from . import utils
 
 __all__ = ['BCycleSystem', 'BCycleStation']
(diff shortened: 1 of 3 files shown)


jasongdove:
Hawaii B-cycle has closed; if there are static lists it should be removed.

mstrcnvs:
The page still exists though, was this information confirmed?

jasongdove:
Yes, from the Hawaii B-cycle page:

The Department of Health Healthy Hawaii Initiative (DOH HHI) has concluded Hawaii B-cycle, the Kailua bikeshare pilot project.  Please visit bikesharehawaii.org to learn more about bikeshare in Hawaii.

mstrcnvs:
Removed Hawaii B-cycle as advised.

mstrcnvs:
Also fixed error handling that was being obfuscated with the previous method. Now Purgatory stations are ignored but other stations will still raise attention if we can't parse them.

eskerda:
LGTM :+1:

eskerda:
> Though we still have a failing test for the Hawaii B-cycle, which for some reason has no stations on their site. What can we do about it so bcycle passes all tests?

That's a problem we are having with tests too on other systems that will occasionally report 0 stations. Can't think of a good solution, as ignoring that a system reports 0 stations will kind of defeat the purpose of the tests themselves. One thing pending though is making the tests a bit more atomic instead of testing lots of things at the same time, thus failing and providing little information on what the problem is.

mstrcnvs:
Good! I'm using PyBikes for a project, including on Python 3, so expect more PRs from me ;)

PR #53, opened 2015-07-28
Adding Encicla bike system from Medellin, Colombia

eduardom:
Hi, it is my first contribution to pybikes. First of all, thanks and congratulations for the great work in the pybikes library. I am planning to use it a lot through the API for a future service I will create. Therefore, I came here to add a new bike system. This is Encicla, from Medellin in Colombia.

Today I've started looking at pybikes' code and learning how to add a new system. Well, few hours later I've added this one, which basically drags data from a feed_url. I've used the bixi as an example.

I am looking forward for the continuation of your blog post (http://eskerda.com/categorizing-citybikes-data-feeds/) in order to get the grips with the different systems. By the way, what the sync variable does?

Thanks.

### DIFF
--- a/pybikes/data/bicincitta.json
+++ b/pybikes/data/bicincitta.json
@@ -17,7 +17,7 @@
                 {
                     "meta": {
                         "latitude": 44.335084,
-                        "city": "Santa Margherita Ligure - Portofino - Camogli",
+                        "city": "Santa MargherIT Ligure - Portofino - Camogli",
                         "name": "Portofino Park and Bike",
                         "longitude": 9.211264,
                         "country": "IT"
@@ -808,7 +808,7 @@
                     "system_id": 65,
                     "meta": {
                         "name": "BICIamo",
-                        "city": "Cantù",
+                        "city": "Cantu'",
                         "country": "IT",
                         "latitude": 45.73626,
                         "longitude": 9.12569
@@ -840,7 +840,7 @@
                     "tag": "ossola",
                     "system_id": 11,
                     "meta": {
-                        "name": "Comunità Montana Valli dell'Ossola",
+                        "name": "Comunita' Montana Valli dell'Ossola",
                         "city": "Santa Maria Maggiore - Malesco - Villette",
                         "country": "IT",
                         "latitude": 46.127196,
--- a/pybikes/data/bicincitta.json
+++ b/pybikes/data/bicincitta.json
@@ -17,7 +17,7 @@
                 {
                     "meta": {
                         "latitude": 44.335084,
-                        "city": "Santa MargherIT Ligure - Portofino - Camogli",
+                        "city": "Santa Margherita Ligure - Portofino - Camogli",
                         "name": "Portofino Park and Bike",
                         "longitude": 9.211264,
                         "country": "IT"
@@ -808,7 +808,7 @@
                     "system_id": 65,
                     "meta": {
                         "name": "BICIamo",
-                        "city": "Cantu'",
+                        "city": "Cantù",
                         "country": "IT",
                         "latitude": 45.73626,
                         "longitude": 9.12569
@@ -840,7 +840,7 @@
                     "tag": "ossola",
                     "system_id": 11,
                     "meta": {
-                        "name": "Comunita' Montana Valli dell'Ossola",
+                        "name": "Comunità Montana Valli dell'Ossola",
                         "city": "Santa Maria Maggiore - Malesco - Villette",
                         "country": "IT",
                         "latitude": 46.127196,
--- a/pybikes/data/encicla.json
+++ b/pybikes/data/encicla.json
@@ -6,7 +6,7 @@
                 "latitude": 6.254942,
                 "longitude": -75.567982,
                 "city": "Medellin",
-                "name": "Encicla",
+                "name": "EnCicla",
                 "country": "CO"
             },
             "feed_url": "http://encicla.gov.co/status"
--- a/pybikes/data/encicla.json
+++ b/pybikes/data/encicla.json
@@ -9,8 +9,7 @@
                 "name": "Encicla",
                 "country": "CO"
             },
-            "feed_url": "http://encicla.gov.co/status", 
-            "format": "json"
+            "feed_url": "http://encicla.gov.co/status"
         }
     ],
     "system": "encicla",
--- a/pybikes/data/encicla.json
+++ b/pybikes/data/encicla.json
@@ -1,0 +1,18 @@
+{
+    "instances": [
+        {
+            "tag": "encicla",
+            "meta": {
+                "latitude": 6.254942,
+                "longitude": -75.567982,
+                "city": "Medellin",
+                "name": "Encicla",
+                "country": "CO"
+            },
+            "feed_url": "http://encicla.gov.co/status", 
+            "format": "json"
+        }
+    ],
+    "system": "encicla",
+    "class": "Encicla"
+}
--- a/pybikes/encicla.py
+++ b/pybikes/encicla.py
@@ -1,0 +1,104 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2015, Eduardo Mucelli Rezende Oliveira <edumucelli@gmail.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import json
+import codecs
+
+from lxml import etree
+
+from .base import BikeShareSystem, BikeShareStation
+from . import utils, exceptions
+
+__all__ = ['Encicla', 'EnciclaStation']
+
+parse_methods = {
+  'json': 'get_json_stations',
+}
+
+class Encicla(BikeShareSystem):
+
+  sync = True
+
+  meta = {
+    'system': 'Encicla',
+    'company': 'Sistema de Bicicletas Públicas del Valle de Aburrá'
+  }
+
+  def __init__(self, tag, feed_url, meta, format):
+    super(Encicla, self).__init__(tag, meta)
+    self.feed_url = feed_url
+    self.method = format
+
+  def update(self, scraper = None):
+    if scraper is None:
+      scraper = utils.PyBikesScraper()
+
+    if self.method not in parse_methods:
+      raise Exception(
+        'Extractor for method %s is not implemented' % self.method )
+
+    self.stations = eval(parse_methods[self.method])(self, scraper)
+
+def get_json_stations(self, scraper):
+  data = json.loads(scraper.request(self.feed_url))
+  stations = []
+  
+  for station in data['stations']:
+    for item in station['items']:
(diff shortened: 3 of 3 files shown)


eskerda:
Reviewing

eskerda:
Hello, thank you for your contribution.

Some things about the PR:

1. It's really important to keep the indentation of the original library. In this case, it's indented using 4 spaces. Your code is indented using 2 spaces.
2. Usually patches should follow the same theme. In this case, fixing the names of cities in Italy, whilst gladly accepted, would usually be done on a separate PR.
3. Each spider will be slightly different to the others. The main idea is to keep them as generic to a system as possible to allow further reuse. If there are no other systems in the world that share the encicla implementation, then it's completely alright to include the feed url inside the spider and not the data file. 
4. Following number 3, the implementation details of bixi differ from encicla, which is simply consuming a json feed. On that case, there's no need to specify many different things. I will add comments to the code.

pybikes/data/encicla.json
```diff
@@ -0,0 +1,18 @@
...
+            "tag": "encicla",
+            "meta": {
+                "latitude": 6.254942,
+                "longitude": -75.567982,
+                "city": "Medellin",
+                "name": "Encicla",
+                "country": "CO"
+            },
+            "feed_url": "http://encicla.gov.co/status", 
+            "format": "json"
```
eskerda: This format `json` was just a formality in bixi to support multiple feeds with an slightly compatible implementation. On this case, not knowing if there are more systems that share this syntax it's safe to just assume it's always going to be a json.

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2015, Eduardo Mucelli Rezende Oliveira <edumucelli@gmail.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import json
+import codecs
```
eskerda: What does import codecs do?

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
...
+from lxml import etree
+
+from .base import BikeShareSystem, BikeShareStation
+from . import utils, exceptions
+
+__all__ = ['Encicla', 'EnciclaStation']
+
+parse_methods = {
+  'json': 'get_json_stations',
+}
```
eskerda: Assuming it's always going to be a json feed, we can also delete this trickery

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
...
+
+  def update(self, scraper = None):
+    if scraper is None:
+      scraper = utils.PyBikesScraper()
+
+    if self.method not in parse_methods:
+      raise Exception(
+        'Extractor for method %s is not implemented' % self.method )
+
+    self.stations = eval(parse_methods[self.method])(self, scraper)
```
eskerda: Same here as before. In bixi this made some sense. Here it would be just abusing the style too much

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get_json_stations(self, scraper):
+  data = json.loads(scraper.request(self.feed_url))
+  stations = []
+  
+  for station in data['stations']:
+    for item in station['items']:
+      station = EnciclaStation.from_json(item)
+      stations.append(station)
```
eskerda: Being a json feed trivial to consume there's no need to create an special EnciclaStation object.

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
...
+      }
+    '''
+    station = EnciclaStation()
+    
+    station.name      = data['name']
+    station.longitude = float(data['lon'])
+    station.latitude  = float(data['lat'])
+    station.bikes     = int(data['bikes'])
+    places            = data['places']
+    if places is None:
```
eskerda: For most cases

```
if places is None
```

is equivalent to

```if not places```

File: pybikes/encicla.py [hunk 7]
```diff
@@ -0,0 +1,104 @@
...
+    else:
+      station.free    = int(places)
+
+    # 'bikes_state' may also be a label as 'warning' as well, do not try to cast it to int
+    station.extra = {
+      'order': int(data['order']),
+      'address': data['address'],
+      'description': data['description'],
+      'type': data['type'],
+      'capacity': data['capacity'],
```
eskerda: This is my fault for not documenting it better, but the accepted way to name `capacity` on pybikes is `slots` in extra

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
...
+    places            = data['places']
+    if places is None:
+      # For some stations that present 'null' for the 'places' attribute
+      station.free    = int(data['capacity']) - int(data['bikes'])
+    else:
+      station.free    = int(places)
+
+    # 'bikes_state' may also be a label as 'warning' as well, do not try to cast it to int
+    station.extra = {
+      'order': int(data['order']),
```
eskerda: What does `order` stand for on the feed? If it's useful it can stay. For fields that will not be useful, we usually just discard them.

pybikes/encicla.py
```diff
@@ -0,0 +1,104 @@
...
+      'address': data['address'],
+      'description': data['description'],
+      'type': data['type'],
+      'capacity': data['capacity'],
+      'picture': data['picture'],
+      'bikes_state': data['bikes_state'],
+      'places_state': data['places_state'],
```
eskerda: I have to guess that bikes state and places state on the feed means if bikes or slots are operating normally?

eduardom:
Hi eskerda,

thanks for the detailed comment. I will soon fix what you've pointed out. I will try to investigate what places null means, but I am not sure I will for sure know it.

eduardom:
In this new PR, I've fixed the aspects mentioned before.
A - Removed the format json in the encicla.json
B - Removed the package codecs, which I've included because I thought it was part of some internals. I did so because it is present in bixi too, then I thought it was used somehow.
C - Removed all trickery related to format_json 
D - Fixed slots in extra
E - As you've guessed, order is just the identifier within the "Comuna", so I guess it will not be important here, removed it from extras.
F - From what I've observed, places null happen in a random way, I could not guess what they really meant. One thing is, they happen mostly in the first and second "Comuna", but it does not helps us here. Another problem I've realized is that 'capacity' field is not reliable, sometimes the 'capacity' is even less than the number of 'bikes'. Instead of guessing the number of free bikes for the stations with places null, I've set then as 0. Even the Encicla interactive map does not show the number of free bikes for such stations. Thus, I preferred not to make a guess myself.

eskerda:
There might be more context on what fields mean at http://encicla.gov.co/wp-content/themes/encicla/js/map.js

For instance, about `places_state` and `bikes_state`, we can see they are just used to declare the class of some info popups on the map page, so we should take them out of the data feed:
http://encicla.gov.co/wp-content/themes/encicla/js/map.js , line 330:
```
if (station.cdo) {
    labels_class += 'labels-' + station.type;
    info_class += 'info-' + station.type;
} else if (!station.closed) {
    labels_class += 'labels-' + station.type + ' labels-state-' + station.bikes_state;
    info_class += 'info-' + station.type + ' info-bikes-state-' + station.bikes_state + ' info-places-state-' + station.places_state;
}
```

Also, about the `cdo`, it probably means `Centro de Operaciones`, as seen on this screenshot. So we should drop them (as they are not stations): 
image.

pybikes/encicla.py
```diff
@@ -0,0 +1,84 @@
...
+        # it is not straightforward to know what it actually means. Consequently, it was not
+        # included in the extras on the 'slots' field.
+        self.extra = {
+            'address': item['address'],
+            'description': item['description'],
+            'type': item['type'],
+            'picture': item['picture'],
+            'bikes_state': item['bikes_state'],
+            'places_state': item['places_state'],
+            'closed': int(item['closed']),
```
eskerda: A boolean conversion here is in order:
```
'closed': bool(item['closed']),
```

eduardom:
Thanks for more those inputs. Last commit address them.

eduardom:
Any update on that one, @eskerda ?

pybikes/encicla.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2015, Eduardo Mucelli Rezende Oliveira <edumucelli@gmail.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import json
+
+from lxml import etree
```
eskerda: Unused import

pybikes/data/encicla.json
```diff
@@ -0,0 +1,17 @@
+{
+    "instances": [
+        {
+            "tag": "encicla",
+            "meta": {
+                "latitude": 6.254942,
+                "longitude": -75.567982,
+                "city": "Medellin",
+                "name": "Encicla",
```
eskerda: I've seen on the website they name it `EnCicla`

pybikes/encicla.py
```diff
@@ -0,0 +1,83 @@
...
+            
+        stations = []
+        
+        data = json.loads(scraper.request(self.feed_url))
+        for station in data['stations']:
+            for item in station['items']:
+                # discard 'Centro de Operaciones' (Operation Center) from the set of stations
+                if not bool(int(item['cdo'])):
+                    station = EnciclaStation(item)
+                    stations.append(station)
```
eskerda: Usually clearer:
```
if int(item['cdo']) != 0:
    continue
station = EnciclaStation(item)
stations.append(station)
```

pybikes/encicla.py
```diff
@@ -0,0 +1,83 @@
...
+        self.bikes     = int(item['bikes'])
+        places         = item['places']
+        if not places:
+            self.free = 0
+        else:
+            self.free  = int(places)
+        # 'bikes_state' may also be a label as 'warning' as well, do not try to cast it to int
+        # 'capacity' field may present a number which is smaller than the number of 'bikes',
+        # it is not straightforward to know what it actually means. Consequently, it was not
+        # included in the extras on the 'slots' field.
```
eskerda: Long comment could be shorter, ie:
```
# 'capacity' is often incorrect, even smaller than the number of bikes
# therefore it was not included on the 'slots' field
```

pybikes/data/bicincitta.json
```diff
@@ -840,7 +840,7 @@
                     "tag": "ossola",
                     "system_id": 11,
                     "meta": {
-                        "name": "Comunita' Montana Valli dell'Ossola",
+                        "name": "Comunità Montana Valli dell'Ossola",
```
eskerda: Can you please include these changes in another PR?

eduardom:
Thanks for more those comments @eskerda. Just commited a version with them.

eskerda:
:+1: Great!

PR #58, opened 2015-08-30
Fix: Bikeu recently changed one of its URLs and map code

eduardom:
This PR fixes the recent changes in Bikeu system. They've basically changed the location of the map in the 'srm' and the map code.


eskerda:
:+1:

PR #59, opened 2015-08-31
Adding two cities to Samba, Fortaleza and Aracajú with necessary code

eduardom:
Mobilicidade system (aka Samba) has two more cities in which the map code varies from the previous ones. In order to deal with that, I've added a new Class SambaNew (following the Bicincitta idea) to deal with its specificities.

Moreover, I've cleaned up the SambaStation code, which now receives the clean parameters. With this change, each of the classes take care of the data and SambaStation only attributes them to the respective variables, plus extra.


pybikes/samba.py
```diff
@@ -32,35 +36,62 @@ def update(self, scraper = None):
...
+        stations = re.findall(Samba._STATIONS_RGX, html_data)
 
         self.stations = []
 
         # The regex will also match for a function defined in the html. This
         # function is in the position of the array, and thus the [:-1]
         for station in stations[:-1]:
-            self.stations.append(SambaStation(station.split(',')))
+            data = station.split(',')
+            name, latitude, longitude, bikes, free, online_status, operation_status = data[3], float(data[0]), float(data[1]), int(data[7]), int(data[8]) - int(data[7]), data[5], data[6]
```
eskerda: Can you please break this line following a normal variable assignment as:

`
name = data[3]
latitude = float(data[0])
...
```

It's really difficult to know which variable is assigned what.

File: pybikes/samba.py [hunk 2]
```diff
@@ -32,35 +36,62 @@ def update(self, scraper = None):
...
+        stations = ast.literal_eval(stations_html[0])
         '''
-        data is a list of strings, in the following order:
-            [latitude, longitude, icon (path to image file, which we ignore),
-             name, stationId, onlineStatus, operationStatus, availabeBikes,
-             bike capacity, address]
+        Different from the original Samba class, th new one deals receives stations' information in the following format:
+        [(0) name, (1) latitude, (2) longitude, (3) address, (4) address main line, (5) onlineStatus, (6) operationStatus, (7) available bikes (variable not being used in their code) 
+            (8) available bikes, (9) available bike stands, (10) internal station status, (11) path to image file, (12) stationId]
         '''
```
eskerda: Please limit lines to 80 characters if possible.

pybikes/samba.py
```diff
@@ -32,35 +36,62 @@ def update(self, scraper = None):
...
+
+    def update(self, scraper = None):
+        if scraper is None:
+            scraper = utils.PyBikesScraper()
+        scraper.setUserAgent(USERAGENT)
+
+        html = scraper.request(self.feed_url)
+
+        stations_html = re.findall(SambaNew._STATIONS_RGX, html)
+        stations = ast.literal_eval(stations_html[0])
```
eskerda: ? Why?

eduardom: As described in the PR original text, I've thought about using the same as Bicincitta, i.e., a "BicincittaOld" and "Bicincitta" classes to deal with the different implementations within the same network.

What should be then the approach?

eskerda: I was referring to `ast.literal_eval`. It works because there's an overlap between js and python syntax of declaring lists and dicts, but I don't like it as a tool for parsing js objects (and is usually not recommended).

eduardom: Looks like ast.literal_eval is actually the way to do, check what Martijn Pieters says: http://stackoverflow.com/a/14347168/914874

eskerda: Well, if Martihn Pieters says so... Actually transforming it for json reading is not a good solution either and literal_eval is assumed to be safe.

eduardom: What would be then the best approach for it? I've thought json.loads would do the trick, but it breaks because each of the array's elements is not properly quoted. I am stuck, @eskerda.

eskerda: It's alright with `literal_eval`. Checking other issues with the PR

eskerda:
Merged, thanks!

PR #60, opened 2015-08-31
Decobike updates

eduardom:
(1) Decobikelbny does not exist anymore.
(2) Decobike Miami has been rebranded to Citi Bike Miami, but still operated by DecoBike.
(3) New feed url for Miami Beach, which contains update stations' information.


eskerda:
:ship:

PR #63, opened 2015-09-05
Fix: Bcycle new Events Marker breaks the current code

eduardom:


eduardom:
Did not knew about InvalidStation exception. Done, @eskerda .

eskerda:
Thanks!

PR #64, opened 2015-09-05
Adding Cyclopolis System

edumucelli:


pybikes/data/cyclopolis.json
```diff
@@ -0,0 +1,127 @@
+{
+    "instances": [
+        {
+            "tag": "cyclopolis-maroussi",
+            "meta": {
+                "latitude": 38.0568722388,
+                "longitude": 23.8083299536,
+                "city": "Maroussi",
+                "name": "Cyclopolis Maroussi",
```
eskerda: Similar to nextbike, please remove name if we do not have any available name, and name just consists of system_name + city_name. Pybikes will return it as 'Cyclopolis', since city information is already under city.

eduardom:
Here it goes, @eskerda.

eskerda:
Looking good :+1: 

Thanks again!

PR #57, opened 2015-08-21
Adding 96 Nextbikes locations

gszathmari:
Adding support for 96 Nexbikes locations in the following cities:
- 10vorWien
- Al Sharjah
- Auckland
- Augsburg
- Bath
- Berlin
- Bielefeld
- Bietigheim-Bissingen
- Bochum
- Bottrop
- Cambridge
- Christchurch
- Dobrich
- Dortmund
- Dresden
- Dubai
- Duisburg
- Düsseldorf
- Essen
- Flensburg
- Frankfurt
- Friedrichshafen
- Gelsenkirchen
- Glasgow
- Gospić
- Grodzisk Mazowiecki
- Gütersloh
- Hamburg
- Hamm
- Hannover
- Heidelberg
- Herne
- Herrenberg
- Hoboken
- Hollabrunn
- Innsbruck
- Juchnowiec Kościelny
- Jūrmala
- Karlovac
- Karlsruhe
- Kent State University
- King Abdullah Economic City
- Konya
- Krems Ost
- Laa an der Thaya
- Leipzig
- Limassol
- Lublin
- Ludwigsburg
- Ludwigshafen
- Luzern
- Magdeburg
- Mannheim
- Marchfeld
- Mödling
- Mülheim a.d.R.
- München
- NeusiedlerSee
- Norderstedt
- Nürnberg
- ÖBB-Bahnhöfe
- Oberes Ybbstal
- Oberhausen
- Offenbach am Main
- Offenburg
- Piestingtal
- Pittsburgh
- Potsdam
- Quickborn
- Rīga
- Römerland
- Schwieberdingen
- Seferihisar
- Serfaus
- Šibenik
- Speyer
- Stirling
- St.Pölten
- Südheide
- Sursee
- Thermenregion
- Traisen-Gölsental
- Triestingtal
- Tübingen
- Tulln
- Tullnerfeld West
- Unteres Traisental
- Usedom
- Vaihingen an der Enz
- Wachau
- Waiblingen
- West Palm Beach Florida
- WienerWald
- Wr.Neustadt
- Würzburg
- Zagreb

PR #66, opened 2015-09-16
added meta.names for gszathmari's 96 nextbike locations

kawie:
after reading the discussion in eskerda/pybikes#57 I manually edited the `meta.name` tag to reflect pybikes naming conventions. 

in detail:
- removed `meta.name` for „nextbike [Country]“
- removed all city and country names from `meta.name`, except where part of the branding
- also, nextbike-gospic changed domain from hr to gs. so I changed that.

that last point is kind of critical though, because if nextbike decides to change another domain at some point in the future, support for that city in pybikes will probably break. might want to look into that.

I ran the tests for nextbike (took forever), everything seems to be working a-okay.

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -1,25 +1,13 @@
 {
     "instances": [
         {
-            "domain": "bu",
-            "tag": "belfast-bikes",
-            "meta": {
-                "latitude": "54.59488",
-                "city": "Belfast",
-                "name": "Belfast Bikes",
-                "longitude": "-5.9266323",
-                "country": "UK"
-            },
-            "city_uid": "238"
-        },
-        {
             "domain": "de",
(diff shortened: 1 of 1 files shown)


eskerda:
Looks good. Thanks for the shared effort @gszathmari and @kawie!

gszathmari:
Thanks @eskerda and @kawie

eskerda:
> * also, nextbike-gospic changed domain from hr to gs. so I changed that.
> 
> that last point is kind of critical though, because if nextbike decides to change another domain at some point in the future, support for that city in pybikes will probably break. might want to look into that.

@kawie: as you noted tests take forever to run (since TestSystems downloads info). Safest would be to implement something server side that displays errors per network. Currently jobs that fail are moved to a fail queue, so it's at least feasible to dogfeed some solution. Will think about it

PR #71, opened 2015-10-02
Adding missing bikeu location

javnik36:
Bydgoski rower aglomeracyjny is added (https://www.bra.org.pl/). I'm not able to test it, but seems ok(website js looks the same as another bikeu systems).

*I'm not so familiar with pull request's, so instruct me if something is wrong.*


eskerda:
All looks perfect. Thanks! :+1:

PR #77, opened 2015-10-20
Adding Smoove System

edumucelli:


pybikes/smoove.py
```diff
@@ -0,0 +1,55 @@
...
+
+class SmooveStation(BikeShareStation):
+    def __init__(self, name, latitude, longitude, bikes, free):
+        super(SmooveStation, self).__init__()
+
+        self.name      = name
+        self.latitude  = latitude
+        self.longitude = longitude
+        self.bikes     = bikes
+        self.free      = free
```
eskerda: I think we should modify base BikeShareStation to accept these parameters, defaulting to None. WDYT?

pybikes/smoove.py
```diff
@@ -0,0 +1,55 @@
...
+
+from .base import BikeShareSystem, BikeShareStation
+from . import utils, exceptions
+
+__all__ = ['Smoove', 'SmooveStation']
+
+# Each station is formatted as:
+# newmark_01(28, 45.770958,3.073871, "<div class=\"mapbal\" align=\"left\">022 Jaurès<br>Vélos disponibles: 4<br>Emplacements libres: 10<br>CB: Non<br></div>");
+# I.e., (station_id, latitude, longitude, name, available_bikes, free_bike_stands, credit_card_enabled (discarded afterwards))
+DATA_RGX = r'\(\d+\,\ (\d+.\d+).*?(\d+.\d+)\,\ "<div.*?>(.*?)<br>.*?:(.*?)<br>.*?:(.*?)<br>.*?:(.*?)<br></div>"\)'
```
eskerda: I would cut the regex to get station_id, latitude, longitude, and following as HTML to parse with a selector.

pybikes/smoove.py
edumucelli: That'd be nice. However, it tends to be more my style of writing the parsers to do everything that is necessary in the update method and give clean parameters to the Station class.

We would end up with cleaner station class. I'd go for it.

pybikes/smoove.py
edumucelli: Ok, gonna work on that.

edumucelli: I did that, looks uglier than with regex only because the html that describes the station status is not structure, it is only <br> (not even terminated one) separated.

Take a look.

eskerda: You are right. Then, to make it clearer, let's break it into two composed regexes, so if anything breaks we know into which part look.

```
STATIONS_RGX = r'newmark_\d+\(\s*(\d+)\s*,\s*(\d+.\d+),\s*(\d+.\d+)\s*,\s*\"{}\"'
STATION_RGX = r'<div.*?>(.*?)<br>.*?:(.*?)<br>.*?:(.*?)<br>.*?:(.*?)<br></div>'
DATA_RGX = STATUS_RGX.format(STATUS_RGX)

stations_data = re.findall(DATA_RGX, html)
for uid, lat, lng, name, bikes, free, _ in stations_data:
    ...
    extra = {
        'uid': uid,
    }
...
```

edumucelli:
Done, @eskerda

eskerda:
Thanks! This merge marks a new milestone of 400 networks supported :trophy:

edumucelli:
Oh, great!

PR #85, opened 2015-10-31
[BCycle] adding new cities

edumucelli:
Some new cities for BCycle, Fargo, Battle Creek, Dayton and Philadelphia.


eskerda:
:+1:

PR #86, opened 2015-10-31
[Base] Changing the default BikeSharingStation constructor

edumucelli:
As discussed in #77, the addition of name, lat, lng, bikes, free and extra to the constructor of the BikeSharingStation.

However, the following system classes:

```python
super(SmartBikeStation, self).__init__(0)
super(NextbikeStation, self).__init__(0)
super(CicloSampaStation, self).__init__(0)
super(KeolisStation, self).__init__(0)
super(KeolisStation_v2, self).__init__(index)
```
passed `0` (or `index`) as the `id` parameter. What is this `id` parameter for, that I could not find?

Also, I've cleaned up the `datetime`, as it is never used, I've attributed it directly, instead of checking if it was `None`.

Consequently, I've changed those classes accordingly to remove the attribute.  

Let me know what you think. We can work more on that.

### DIFF
--- a/pybikes/base.py
+++ b/pybikes/base.py
@@ -26,6 +26,8 @@ class BikeShareStation(object):
 
     def __init__(self, name = None, latitude = None, longitude = None,
                        bikes = None, free = None, extra = None):
+        if extra is None:
+            extra = {}
         self.name = name
         self.latitude = latitude
         self.longitude = longitude
--- a/pybikes/base.py
+++ b/pybikes/base.py
@@ -26,15 +26,13 @@ class BikeShareStation(object):
 
     def __init__(self, name = None, latitude = None, longitude = None,
                        bikes = None, free = None, extra = None):
-        if extra is None:
-            extra = {}
         self.name = name
         self.latitude = latitude
         self.longitude = longitude
         self.bikes = bikes
         self.free = free
         self.timestamp = datetime.utcnow()     # Store timestamp in UTC!
-        self.extra = extra
+        self.extra = extra or {}
 
     def __str__(self):
         return "--- {0} ---\n"\
--- a/pybikes/base.py
+++ b/pybikes/base.py
@@ -24,17 +24,15 @@ class BikeShareStation(object):
             - JCDecauxStation, ClearChannelStation
     """
 
-    def __init__(self, id = 0, timestamp = None):
-        if timestamp is None:
-            timestamp = datetime.utcnow()
-
-        self.name = None
-        self.latitude = None
-        self.longitude = None
-        self.bikes = None
-        self.free = None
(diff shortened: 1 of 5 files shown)


eskerda:
> However, the following system classes:
>
>
> ```python
super(SmartBikeStation, self).__init__(0)
super(NextbikeStation, self).__init__(0)
super(CicloSampaStation, self).__init__(0)
super(KeolisStation, self).__init__(0)
super(KeolisStation_v2, self).__init__(index)
```
> passed 0 (or index) as the id parameter. What is this id parameter for, that I could not find?

Long ago an index was being passed. Since this was incremental we got rid of it, but some code never changed. You can ignore it.

> Also, I've cleaned up the datetime, as it is never used, I've attributed it directory, instead of checking if it was None.

Good. The idea was to be able to pass a datetime str manually, but since this was never used, it makes total sense to just default on initialization.

Fun python fact that explains why `None` was being checked on datetime, and why `extra` can't default to a `dict`.

```python
from datetime import datetime
from time import sleep

class Foo(object):

    uber_extra = {}

    def __init__(self, date=datetime.utcnow(), extra={}):
        self.date = date
        self.extra = extra

foo = Foo()
sleep(1)
bar = Foo()
foo.extra['Foo'] = 'foo'
foo.uber_extra['SuperFoo'] = 'super_foo'

bar.extra['Bar'] = 'bar'
bar.uber_extra['SuperBar'] = 'super_bar'

print foo.date
print foo.extra
print foo.uber_extra

print bar.date
print bar.extra
print bar.uber_extra
```

```
~ ❯ python2 python_fun.py
2015-11-01 16:59:16.287535
{'Foo': 'foo', 'Bar': 'bar'}
{'SuperBar': 'super_bar', 'SuperFoo': 'super_foo'}
2015-11-01 16:59:16.287535
{'Foo': 'foo', 'Bar': 'bar'}
{'SuperBar': 'super_bar', 'SuperFoo': 'super_foo'}
```

edumucelli:
I see, @eskerda, true! Good catch! Committed accordingly.

pybikes/base.py
```diff
@@ -24,17 +24,15 @@ class BikeShareStation(object):
...
-        self.extra = {}
+    def __init__(self, name = None, latitude = None, longitude = None,
+                       bikes = None, free = None, extra = None):
+        self.name = name
+        self.latitude = latitude
+        self.longitude = longitude
+        self.bikes = bikes
+        self.free = free
+        self.timestamp = datetime.utcnow()     # Store timestamp in UTC!
+        self.extra = extra
```
eskerda: Let's default extra to a `dict`: 
```
self.extra = extra or {}
```

eskerda:
Looks good. Thanks!

PR #87, opened 2015-11-04
Add Velobike station based bikesharing Astana

BenSto:
I have seen this Provider in Astana, Kazachstan. Their Webservice is unfortunately very slow, so I dont know if this is a candidate for the System. Maybe you can have a look by yourself @eskerda . Thanks in advance!

### DIFF
--- a/pybikes/data/velobike.json
+++ b/pybikes/data/velobike.json
@@ -1,0 +1,17 @@
+{
+    "instances": [
+        {
+            "tag": "velobike",
+            "meta": {
+                "city": "Astana",
+                "name": "Velobike",
+                "country": "KZ",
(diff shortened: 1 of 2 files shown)


eskerda:
Checking...

BenSto:
> Checking...

seems like there is something like a winter break in astana, maybe next spring they start the service again... :disappointed:

eskerda:
Weird, last time I checked the URL endpoint was not working.

System added, sorry for the delay. Hope next spring it can be useful to you

BenSto:
Yeah I think their webservice is not very stable, maybe that was the problem. Thanks for merging! :bike:

PR #88, opened 2015-11-13
[Nextbike] fix Wien name

edumucelli:


eskerda:
Seems like 10 vor wien is a commune in Wien.

PR #89, opened 2015-11-17
Changing GB to ISO_3166-2:GB relative codes

edumucelli:
Scotland, and England are countries and have their own ISO_3166 codes, which are described here: https://en.wikipedia.org/wiki/ISO_3166-2:GB

This PR changes the "country" attribute of those countries accordingly.

### DIFF
--- a/pybikes/data/bicincitta.json
+++ b/pybikes/data/bicincitta.json
@@ -262,7 +262,7 @@
                         "city": "Bath",
                         "name": "Bike in Bath",
                         "longitude": -2.362404,
-                        "country": "GB"
+                        "country": "GB-ENG"
                     },
                     "tag": "bike-in-bath",
                     "endpoint": "http://www.bikeinbath.com/frmLeStazioni.aspx"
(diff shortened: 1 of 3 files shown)


edumucelli:
@eskerda, what about this one here?

eskerda:
Need to think about it. I will take a wild guess and assume some clients are relying on these codes to set up tiny png flags on their apps

edumucelli:
That is not a wild guess, I use this approach for this very feature, flags. Thus, showing "GB" flag for Scotland, Northern Ireland, Wales is just plain wrong. That was the objective of this PR, if someone relies on that for the flag, s/he is going to show the wrong one, as I did.

bcaller:
The GB flag _is_ correct for England, Scotland, Northern Ireland and Wales as all are countries within the United Kingdom (which is also a country).
My wild guess is that most people expect a two-letter ISO 3166-1 country code and may rely on that. With this change they would need to know that they must make a substring.

eskerda:
Flags aside, when I added country as a code it was meant for just providing a way to group networks without having to do too much reverse geocoding.

What about including a `state` field?

PR #95, opened 2015-12-09
Adding Cocacola system

edumucelli:


eskerda:
Even if the system is sponsored by Coca-Cola, seems [Telfourth](http://telfourth.com/) (as [An Rothar Nua](http://www.anrotharnua.ie/)) is the developer and operator. Since they do not provide a general name for the system I think the safest name for the system class will be `BikeshareIE`, filenamed `bikeshare_ie.py`. As per naming, I guess we can name these `Coca-Cola Zero® Bikes`, as much as I hate it, it's their name.

pybikes/data/cocacola.json
```diff
@@ -0,0 +1,33 @@
...
+            "tag": "limerick",
+            "meta": {
+                "latitude": 52.668746,
+                "longitude": -8.627809,
+                "city": "Limerick",
+                "country": "IE"
+            }
+        },
+        {
+            "tag": "galway",
```
eskerda: I don't like tying the tag of the system with the functionality of the implementation. Please use something else, like `system_tag` or `site_tag`.

:+1: on the tags with the city name, this is the safest assumption.

pybikes/cocacola.py
```diff
@@ -0,0 +1,60 @@
...
+class CocaColaStation(BikeShareStation):
+    def __init__(self, name, latitude, longitude, bikes, free, extra):
+        super(CocaColaStation, self).__init__()
+
+        self.name      = name
+        self.latitude  = latitude
+        self.longitude = longitude
+        self.bikes     = bikes
+        self.free      = free
+        self.extra     = extra
```
eskerda: Remember that PR you did with a new constructor method for BikeShareStation?

pybikes/cocacola.py
```diff
@@ -0,0 +1,60 @@
...
+
+        for item in data[self.tag]:
+            name = item['name']
+            latitude = item['latitude']
+            longitude = item['longitude']
+            bikes = item['bikesAvailable']
+            free = item['docksAvailable']
+            extra = {
+                'uid': item['stationId']
+            }
```
eskerda: Don't assume latitude and longitude are floats or bikes and free are ints, do an explicit type cast.

pybikes/cocacola.py
```diff
@@ -0,0 +1,60 @@
...
+        data = json.loads(stations_html[0])
+
+        for item in data[self.tag]:
+            name = item['name']
+            latitude = item['latitude']
+            longitude = item['longitude']
+            bikes = item['bikesAvailable']
+            free = item['docksAvailable']
+            extra = {
+                'uid': item['stationId']
```
eskerda: Add total slots `item['docksCount']` as `slots` on extra

edumucelli:
Just pinging that I will proceed on this soon, just in a shortage of time these days.

pybikes/data/cocacola.json
edumucelli: I will set it as system_id, as on bicincitta. It would be nice if we could "standardize" the name of those attributes too.

edumucelli: Did not get very well, the name part, where should it then be "Coca-Cola Zero® Bikes", the 'company' attribute inside the meta hash?

edumucelli:
Updated, @eskerda

pybikes/data/cocacola.json
eskerda: You can either set it on the meta as `system` in `bikeshare_ie.py` or as name in `data/bikeshare_ie.json`.

eskerda:
@edumucelli 
> Did not get very well, the name part, where should it then be "Coca-Cola Zero® Bikes", the 'company' attribute inside the meta hash?

You can either set it on the meta as system in bikeshare_ie.py or as name in data/bikeshare_ie.json.

edumucelli:
@eskerda what do you think?

eskerda:
:+1: Looking good, awesome!

PR #96, opened 2015-12-09
Velobike company

edumucelli:
Almost in the bottom of their page it is possible to see the logs of the partners.

### DIFF
--- a/pybikes/velobike.py
+++ b/pybikes/velobike.py
@@ -4,16 +4,14 @@ import json
 from .base import BikeShareSystem, BikeShareStation
 from . import utils
 
+
 class Velobike(BikeShareSystem):
 
     sync = True
 
     meta = {
         'system': 'Velobike',
-        'company': ['Agency for Physical Culture and Sports of \
-                    the Republic of Kazakhstan', 
-                    'Sovereign Wealth Fund "Samruk-Kazyna" JSC',
-                    'Akimat of Astana', 'Smoove']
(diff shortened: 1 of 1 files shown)


pybikes/velobike.py
```diff
@@ -4,14 +4,16 @@
 from .base import BikeShareSystem, BikeShareStation
 from . import utils
 
-
 class Velobike(BikeShareSystem):
```
eskerda: Don't delete that line:

> Surround top-level function and class definitions with two blank lines.
> from: https://www.python.org/dev/peps/pep-0008/#blank-lines

pybikes/velobike.py
```diff
@@ -4,14 +4,16 @@
...
 
     sync = True
 
     meta = {
         'system': 'Velobike',
-        'company': 'Velobike.kz, LLP'
+        'company': ['Agency for Physical Culture and Sports of \
+                    the Republic of Kazakhstan', 
+                    'Sovereign Wealth Fund "Samruk-Kazyna" JSC',
+                    'Akimat of Astana', 'Smoove']
```
eskerda: I think we should have `Velobike.kz, LLP` in there too (http://velobike.kz/lease_agreement_en.pdf).

For me, clearest would be:
```
['Velobike.kz, LLP', 'Smoove']
```

edumucelli:
I think would be more complete with the company information as before. But anyway, changed accordingly.

PR #101, opened 2015-12-18
[Smartbike] fix recent change on bikemi

edumucelli:


eskerda:
Awesome work, :ship: it!

PR #102, opened 2016-01-08
Add Clujbike, Romania

BenSto:
Feel free to review. This is a Bikesharing System in Cluj (Romania), which is currently in winter break.
Thanks.

### DIFF
--- a/pybikes/clujbike.py
+++ b/pybikes/clujbike.py
@@ -1,0 +1,95 @@
+# -*- coding: utf-8 -*-
+import json
+
+from .base import BikeShareSystem, BikeShareStation
+from . import utils, exceptions
+
+__all__ = ['Clujbike', 'ClujbikeStation']
+
+class Clujbike(BikeShareSystem):
+
+    sync = True
+
+    meta = {
+        'system': 'Clujbike',
+        'company': 'Municipiul Cluj-Napoca'
+    }
+
+    def __init__(self, tag, feed_url, meta):
+        super(Clujbike, self).__init__(tag, meta)
+        self.feed_url = feed_url
+
+    def update(self, scraper = None):
+        if scraper is None:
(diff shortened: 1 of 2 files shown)


eskerda:
> dont know exactly if this makes sense, to skip the stations that are currently offline.

I think it does not. These are present on their map, and the other status field `Status` says `Functionala`, so it looks like it just represents station having 0 bikes? 

There's one station that we must skip, though, the last one on the feed named `Statie Virtuala` and marked as `Latitude: 0.0000` `Longitude: 0.00000`.

pybikes/clujbike.py
```diff
@@ -0,0 +1,95 @@
...
+class ClujbikeStation(BikeShareStation):
+    def __init__(self, name, latitude, longitude, bikes, free, extra):
+        super(ClujbikeStation, self).__init__()
+
+        self.name       = name
+        self.latitude   = float(latitude)
+        self.longitude  = float(longitude)
+        self.bikes      = int(bikes)
+        self.free       = int(free)
+        self.extra      = extra
```
eskerda: There's no need anymore to build a custom station class if it's not necessarily needed. BikeShareStation constructor accepts name, latitude, longitude, etc parameters.

float and int casts must be done outside the constructor, though.

BenSto:
> There's one station that we must skip, though, the last one on the feed named Statie Virtuala and marked as Latitude: 0.0000 Longitude: 0.00000.

Good Spotting :+1:  . But there is a another Station beside 'Statie Virtuala' called 'Statie Training' which has also lat / lon 0.0 . So I guess the best way to skip these both is to check for lat or lon and not for their names. Im going to commit it this way. If you have an idea, to get that more smarter, just give me another comment :) . Thanks!

eskerda:
Looks good, thanks a lot!

PR #103, opened 2016-01-08
Latitude and Longitude were transposed

BenSto:
Feel free to have a look. Lat and Lon were transposed for VelobikeRU. As you can easily see on the map on: http://api.citybik.es/v2/ 
Thanks


eskerda:
That's no good. Thanks for noticing!

PR #104, opened 2016-01-08
Add ADBC Bikeshare, Abu Dhabi

BenSto:
Feel free to review. This is a small Bikesharing System in Abu Dhabi. Thank you very much!

### DIFF
--- a/pybikes/adbc.py
+++ b/pybikes/adbc.py
@@ -47,26 +47,17 @@ class Adbcbikeshare(BikeShareSystem):
         # }
         for item in data['stations']:
             name = item['n']
-            latitude = item['la']
-            longitude = item['lo']
-            bikes = item['ba']
-            free = item['da']
+            latitude = float(item['la'])
+            longitude = float(item['lo'])
+            bikes = int(item['ba'])
+            free = int(item['da'])
             extra = {
-                'slots' : int(item['ba']) + int(item['da'])
+                'status' : 'online' if item['st'] == 1 else 'offline',
+                'has_bike_keys': item['bk'] and item['bl'],
+                'uid': str(item['id'])
             }
-            station = AdbcbikeshareStation(name, latitude, longitude,
-                                           bikes, free, extra)
+            station = BikeShareStation(name, latitude, longitude,
+                                       bikes, free, extra)
             stations.append(station)
 
         self.stations = stations
-
-class AdbcbikeshareStation(BikeShareStation):
-    def __init__(self, name, latitude, longitude, bikes, free, extra):
-        super(AdbcbikeshareStation, self).__init__()
-
-        self.name       = name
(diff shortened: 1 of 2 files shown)


pybikes/adbc.py
```diff
@@ -0,0 +1,72 @@
...
+class AdbcbikeshareStation(BikeShareStation):
+    def __init__(self, name, latitude, longitude, bikes, free, extra):
+        super(AdbcbikeshareStation, self).__init__()
+
+        self.name       = name
+        self.latitude   = float(latitude)
+        self.longitude  = float(longitude)
+        self.bikes      = int(bikes)
+        self.free       = int(free)
+        self.extra      = extra
```
eskerda: There's no need anymore to build a custom station class if it's not necessarily needed. BikeShareStation constructor accepts name, latitude, longitude, etc parameters.

float and int casts must be done outside the constructor, though.

eskerda:
From the JS source, this is a mapping of what fields on the feed mean:

```
ed.MapStation = function(stationJson) {
        this.id = stationJson.id;
        this.stationName = stationJson.s;
        this.terminalNumber = stationJson.n;
        this.isManuallyBlocked = stationJson.m;
        this.isSchemeSuspended = stationJson.su;
        this.status = stationJson.st;
        this.isBlocked = stationJson.b;
        this.lastUpdatedTimestamp = stationJson.lu;
        this.hasBikeKeyDispenser = stationJson.bk;
        this.isBikeKeyAvailable = stationJson.bl;
        this.latitude = stationJson.la;
        this.longitude = stationJson.lo;
        this.docksAvailable = stationJson.da;
        this.docksUnavailable = stationJson.dx;
        this.bikesAvailable = stationJson.ba;
        this.bikesUnavailable = stationJson.bx;
        if (stationJson.hvm) {
            this.hvm = stationJson.hvm
        }
    };
```

eskerda:
* It's not clear that `bikesAvailable + docksAvailable == slots`: probably `.da + .dx + .ba + .bx` would actually mean total slots. It's not a mandatory field, so if it's not explicitly there we can skip it too.
* `status: 'online' if item['st'] == 1 else 'offline' `
* As far as we can, it's good to have unique identifiers.
  * `.n (aka terminalNumber)` is giving the same as `.s (aka stationName)`, so not this
  * there's a `.id` that we can use, and add it to extra as `'uid': str(item['id'])`
* There seems to be a "key dispenser" of some kind. This is something someone wanting to build an app just for this schema would appreciate:
  * `'has_bike_keys': item['bk'] and item['bl']`

BenSto:
Thanks a lot for your feedback. I made a commit, regarding your comments.

eskerda:
Thank you!!

:bike: :bike: :bike: :bike: :bike: :bike: :bike: :bike: :bike: :bike: :bike:

Issue #97, opened 2015-12-09
Add / check smoove systems


eskerda:
This page lists all systems provided by smoove: http://www.smoove-bike.com/references

edumucelli:
@eskerda, I've verified all the systems within Smoove. The statuses are the following:

0. France - Annemasse (not implemented, does not provide a map to extract the data: http://projet.smoove.fr/)
1. France - Avignon (works, "velopop" system)
2. France - Belfort (not implemented, does not provide a map to extract the data: http://optymo.fr/velo-libre-service/les-stations)
3. France - Chalon-sur-Saône ("reflex-grandchalon" does not work anymore. The site is not working anymore. As I have not found any news showing that the system has been turned off, I guess we can wait a bit to see if it gets back to work.)
4. France - Clermont-Ferrand (works, "c-velo" system)
5. France - Grenoble (not implemented, does not provide a map to extract the data: http://www.metrovelo.fr/fr/p-carte-des-stations.html)
6. France - Lorient (not implemented, although it provides the number of available bikes for 2 stations, it does not provide the number of free slots: http://www.lorient-velo.fr/sag_vls_stations.html)
7. France - Montpellier (not implemented, seems that data can be extracted from http://tam.cartographie.pro/)
8. France - Nice (works, "velobleu" implemented on Veloway system)
9. France - Saint-Etienne (works, "velivert" system)
10. France - Strasbourg (not implemented, same as Lorient, provides the number of available bikes, but not the free slots: http://www.velhop.strasbourg.eu/sag_vls_stations.html)
11. France - Valence (works, "libelo" system)
12. UK - Slough (It is going to differ from the current implementation, It would be better to add it alone on its on, I will do that: https://www.cycleslough.com/hire/location/)
13. Georgia - Batumi (works, "batumvelo" system)
14. Greece - Corfou ("The city of Kerkyra, on Corfou Island, has planed to  set up its fisrt bke-sharing system in September 2010; Unfortunately, the project failed.")
15. Italia - Vicenza ("Smoove has equipped the city of Vicenza in Italy of a demonstration station in September 2011. 
Unfortunately the project has stopped")
16. Kazakhstan - Astana (works, "velobike" system implemented on its on)
17. Russia - Moscow (works, "velobike_ru" system implemented on its on)
18. Thailand - Bangkok (not implemented, there is no interactive map to get data from: http://www.cubike-chula.com/Mains/Map.aspx)
[...]

PR #116, opened 2016-02-06
Adding Cycle Hire system

edumucelli:
It relates to the issue #97.

### DIFF
--- a/pybikes/data/cyclehire.json
+++ b/pybikes/data/cyclehire.json
@@ -12,6 +12,6 @@
             "feed_url": "https://www.cycleslough.com/hire/location/"
         }
     ],
-    "system": "encicla",
-    "class": "Encicla"
+    "system": "cyclehire",
+    "class": "CycleHire"
 }
--- a/pybikes/data/cyclehire.json
+++ b/pybikes/data/cyclehire.json
@@ -1,7 +1,7 @@
 {
     "instances": [
         {
-            "tag": "cyclehire",
+            "tag": "cycle-hire-slough",
             "meta": {
                 "latitude": 51.51135,
                 "longitude": -0.591562,
--- a/pybikes/data/cyclehire.json
+++ b/pybikes/data/cyclehire.json
@@ -1,0 +1,17 @@
+{
+    "instances": [
+        {
+            "tag": "cyclehire",
+            "meta": {
+                "latitude": 51.51135,
+                "longitude": -0.591562,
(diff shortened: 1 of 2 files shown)


eskerda:
```
Testing cyclehire
=============================
Testing Cycle Hire
E
======================================================================
ERROR: test_cyclehire (__main__.TestSystems)
Traceback (most recent call last):
  File "tests/unittest_pybikes.py", line 268, in test_schema
    self._test_systems(schema)
  File "tests/unittest_pybikes.py", line 21, in _test_systems
    self._test_system(instance['tag'], key)
  File "tests/unittest_pybikes.py", line 30, in _test_system
    self._test_update(p_sys)
  File "tests/unittest_pybikes.py", line 78, in _test_update
    instance.update()
  File "build/bdist.linux-x86_64/egg/pybikes/cyclehire.py", line 47, in update
    tree = lxml.html.fromstring(raw_status)
AttributeError: 'module' object has no attribute 'html'

Ran 1 test in 0.293s

FAILED (errors=1)
```

I think you have to import html as a module. If you don't want to have a conflict with the local variable you have `html`, you can always `from lxml import html as foobar` or you can even `import lxml.html`.

pybikes/data/cyclehire.json
```diff
@@ -0,0 +1,17 @@
+{
+    "instances": [
+        {
+            "tag": "cyclehire",
```
eskerda: Let's go for the safest and set the tag as `cycle-hire-slough`. The probability of another "Cycle Hire" popping somewhere is high.

pybikes/data/cyclehire.json
```diff
@@ -0,0 +1,17 @@
...
+            "meta": {
+                "latitude": 51.51135,
+                "longitude": -0.591562,
+                "city": "Slough",
+                "name": "Cycle Hire",
```
eskerda: Their website explicitly say it is called `Cycle Hire Slough`... but we agreed to take the name of the city out of the bike share system, to avoid redundant info like Cycle Hire Slough, Slough, UK. So :+1:

edumucelli:
Done here, @eskerda

eskerda:
Merged at https://github.com/eskerda/pybikes/pull/121

eskerda:
@edumucelli: added a new dependency: demjson. Since there are many cases we are getting data form a JS object, I think it makes sense to have it.

PR #117, opened 2016-02-21
Add Youbike

eskerda:


eskerda:
@edumucelli For this system I used something that could be useful for other systems. Upon a defined geometry file (in this case in kml), spider filters out stations that are not on a set of bounds.

Some systems have test stations on 0.00 or outside the "viewport". Others (like this case) have all the stations together, so there's no real way of knowing which station is on which part.

Do you think it makes sense to generalize it so it would be easier to support on other systems?

edumucelli:
That is a nice piece of code here, congrats. I am wondering if we have this many cases in which such effort (of adding a library that needs to be compiled and generalizing this code on utils.py) worth. I'd say "carry on" because it is certainly a tool that might be handy for other systems.

One doubt I have is: do you build the bounding box based on the coordinates provided by the kml (in this cases), but lets suppose it comes from a more common method, JSON API. In this JSON data some of the coordinates are 0, then, before building the bounding box, that'd be necessary to check if the bounding box presents only valid values. By doing so, the bounding box would still be necessary? Then, what you are saying is that you think of abstracting all the invalid cases we have today inside the the generalized form of this method on utils.py? Or that by using a bounding box only it would deal with all the invalid coordinates we have so far?

Would be nice to compile all the situations we have in which the latitude and longitude are invalid, besides 0 and "out of bounds".

Issue #128, opened 2016-04-18
Add Helsinki


eskerda:
Possible endpoints:

- Smoove API: https://helsinki-fi-smoove.klervi.net/api-public/stations That one is still missing the data, but the format should be same as in the demo here: https://demo-v2.fr.smoove.klervi.net/api-public/stations
- OpenTripPlanner API (real list of stations, realtime status will be fetched from Smoove): http://beta.digitransit.fi/otp/routers/hsl/bike_rental

tuukka:
Smoove API endpoint has changed at the last minute, and some of the first stations have appeared there already: http://helsinki-fi.smoove.pro/api-public/stations

tuukka:
This system and both APIs are now operational.

eskerda:
@tuukka 

We went for the Smoove feed, since it has total slots value and station id was easy to extract from the name. Hopefully, more smoove feeds will move to a public API, so we could reuse this code too.

Find an online feed at: http://api.citybik.es/v2/networks/citybikefinland-hsk
Most apps supporting the CityBikes API will now support the system too

A couple of questions:
- Is CityBikeFinland the correct name for the system?
- I added the following as companies holding the system. Is it correct?
```
      "Helsinki City Transport", 
      "Helsinki Regional Transport Authority", 
      "CityBikeFinland", 
      "Smoove SAS", 
      "Moventia"
```

tuukka:
@eskerda Great to have you and all your API users onboard, thank you! CityBike Finland is the company that operates the system. I don't think the system has an official name, which is why the parties refer to it by many names. "Helsinki city bikes" might be clearest.

I got the recommendation from Smoove to change to using their newer API endpoint URL: http://helsinki-fi.smoove.pro/api-public/stations

In the response, `style: "CB"` means that the station has a payment card terminal for registering to the system.

tuukka:
Regarding the list of companies, it's good as it is. If you wanted to shorten it, you could leave out Smoove and Moventia, as CityBike Finland is the company owned by them as a joint venture.

eskerda:
@hugovk fixed the issue with the naming. I updated the tag too, so current feed is http://api.citybik.es/v2/networks/citybikes-helsinki and http://api.citybik.es/v2/networks/citybikefinland-hsk redirects to the new one too, in case someone was already reading from there.

Going to add the `cb` as an extra and we are good! Thanks @tuukka

PR #131, opened 2016-05-04
Update Helsinki name

hugovk:
I've not seen the scheme called CityBikeFinland anywhere, they're just called "city bikes" or "Helsinki city bikes" in English:
https://www.hsl.fi/en/citybikes

Also add the sponsoring company.

Re: https://github.com/eskerda/pybikes/issues/128#issuecomment-216623189

### DIFF
--- a/pybikes/data/smoove.json
+++ b/pybikes/data/smoove.json
@@ -7,14 +7,15 @@
                     "tag":"citybikefinland-hsk",
                     "meta":{
                         "city":"Helsinki",
-                        "name":"CityBikeFinland",
+                        "name":"City bikes",
                         "country":"FI",
                         "company":[
                             "Helsinki City Transport",
-                            "Helsinki Regional Transport Authority",
+                            "Helsinki Regional Transport",
                             "CityBikeFinland",
                             "Smoove SAS",
-                            "Moventia"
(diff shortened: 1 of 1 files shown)


eskerda:
Thanks @hugovk! I tried but never managed to reach that previous link.

Current tag of the system is `citybikefinland-hsk`. I can move it to `citybikes-helsinki` too

hugovk:
You're welcome!

By the way there's another API endpoint that has more-or-less the same info:

https://dev.hsl.fi/matka.hsl.fi/otp/routers/hsl/bike_rental

Also JSON: 

`curl -sH "Accept: application/json" https://dev.hsl.fi/matka.hsl.fi/otp/routers/hsl/bike_rental`

eskerda:
> By the way there's another API endpoint that has more-or-less the same info:

We went for the smoove feed, in hopes we can reuse it for other systems. Plus it has bank card information and total slots.

On the other side, I see OTP has [BikeRentalStation](http://dev.opentripplanner.org/apidoc/0.15.0/el_ns0_BikeRentalStationList.html) documented so it's probably a good idea to implement that one. Feeding from an OTP feed provided by HSL is preferable than going for the smoove feed.

Issue #134, opened 2016-05-10
"InsecureRequestWarning: Unverified HTTPS request is being made."


hugovk:
pybikes gives an InsecureRequestWarning for Python 2.7.9 and Python 2.7.11 on Windows 7:

> InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. 

https://urllib3.readthedocs.io/en/latest/security.html#insecurerequestwarning

For example:

```python
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:40:30) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pybikes
>>> bikeshare = pybikes.get('citybikes-helsinki')
>>> bikeshare.update()
C:\Python27\lib\site-packages\requests\packages\urllib3\connectionpool.py:791: InsecureRequestWarning: Unverified HTTPS
request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/late
st/security.html
  InsecureRequestWarning)
>>>
```

eskerda:
Not all networks supported have valid SSL certificates (...)

Maybe we should do it the other way around. Set SSL verification by default, and disable it manually from spider.

eskerda:
Found it. It is not a network, but a failure when using an https proxy:

https://github.com/eskerda/pybikes/commit/b1302a1518cb4d15fb88889dcbd310d4dc31c153

hugovk:
Thanks, that's silenced it.

eskerda:
Seems affecting network is veloway

```
In [1]: import pybikes
In [2]: scraper = pybikes.utils.PyBikesScraper()
In [3]: scraper.request('http://www.dkbus.com/cartoV3/libProxyCarto.asp')
SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
```

hugovk:
Actually, I re-tested incorrectly and it didn't silence it.

hugovk:
Also, with latest master (Python 2.7.11, Win 7):
```python
>>> scraper = pybikes.utils.PyBikesScraper()
>>> scraper.request('http://www.dkbus.com/cartoV3/libProxyCarto.asp')
C:\Python27\lib\site-packages\requests\packages\urllib3\connectionpool.py:821: InsecureRequestWarning: Unverified HTTPS
[...]

PR #136, opened 2016-05-10
Fix dkvelo bike share system

eskerda:


eskerda:
Fixes dkvelo failing #134

PR #149, opened 2016-06-08
Bikeu systems update

thesebas:
Fixed regexp and urls.


eskerda:
See improved changes at https://github.com/eskerda/pybikes/pull/152

PR #152, opened 2016-06-11
Improve bikeu

eskerda:
Hello @thesebas , I've added some improvements over your changes. One of the networks was failing, since the map loads inside as an iframe that requires a token.

Also added bike uid information on extra too. Maybe some people will find it interesting to track bike movements between stations.


thesebas:
You are reading my mind, both fixing this iframe and parsing extra bike data was on my todo list... later. 😊 Is this `bike_uids` extra field standarised?

eskerda:
> Is this bike_uids extra field standarised?

Kind of, nextbike and punpunbikeshare systems do have it.

PR #154, opened 2016-06-27
Add MVG Rad

robbi5:
This PR adds MVG Rad (not to be confused with "MVG Mein Rad" from the MVG _Mainz_ - this is from the MVG _Munich_).
It is basically a white-label nextbike instance, but the full data only appears on `mvgrad.nextbike.net` - so I tried to add it as subclass of nextbike.

### DIFF
--- a/pybikes/data/mvgrad.json
+++ b/pybikes/data/mvgrad.json
@@ -1,0 +1,17 @@
+{
+    "instances": [
+        {
+            "tag": "mvgrad",
+            "meta": {
+                "city": "München",
+                "name": "MVG Rad",
+                "country": "DE",
+                "longitude": 11.575555,
+                "latitude": 48.137222
+            },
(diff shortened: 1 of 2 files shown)


eskerda:
> It is basically a white-label nextbike instance, but the full data only appears on mvgrad.nextbike.net - so I tried to add it as subclass of nextbike.

In this case it's not such a great idea. It's replicating and adding a different behavior on this particular network, but the feed is exactly the same.

I think it's better if we add support for a "hostname" field on the feed at `data/nextbike.json`, defaulting to the current set one at nextbike.net. Also, it's possible this will happen with more networks, so it's good to add this to base nextbike class.

About approximating bike numbers by getting `bike_types`, sure, if that works out for all nextbike feeds sounds like a good idea.

robbi5:
replaced by #164.

Issue #148, opened 2016-06-07
Add readybike bike share system (Reading)


eskerda:
https://www.readybike.co.uk/

edumucelli:
@eskerda, work in progress ...

PR #156, opened 2016-06-28
Add Hourbike (Reading, Liverpool, Liconln, Sheffield) #148

edumucelli:
In response to #148. It is important to know, though, that the Hourbike system uses the same BikeU system for two cities: Oxford (Oxonbikes) and Northampton (Cycle Connect):
- [Oxonbikes](https://www.oxonbikes.co.uk/map-of-stations.html)
- [Cycle Connect](https://www.cycleconnect.co.uk/map-of-stations.html)

As the systems (Hourbike and BikeU) are managed by different entities, I think the correct approach here is to replicate the code from BikeU inside Hourbike (using a second class, e.g., HourbikeAlternative). That would be better than the obvious merging of Hourbike cities into BikeU and replicating this code here inside BikeU. I suggest that because if one of those companies change the system, we are screwed and will have to split the code again.

@eskerda, WDYT?

### DIFF
--- a/pybikes/data/hourbike.json
+++ b/pybikes/data/hourbike.json
@@ -1,0 +1,50 @@
+{
+    "instances": [
+        {
+            "tag": "readybike",
+            "feed_url": "https://www.readybike.co.uk/map-of-stations.html",
+            "meta": {
+                "latitude": 51.434332,
+                "longitude": -0.96866,
+                "city": "Reading",
+                "name": "Readybike",
+                "country": "GB"
+            }
+        },
+        {
+            "tag": "citybike-liverpool",
+            "feed_url": "https://www.citybikeliverpool.co.uk/map-of-stations.html",
+            "meta": {
+                "latitude": 53.358322,
+                "longitude": -2.864017,
+                "city": "Liverpool",
+                "name": "Citybike Liverpool",
+                "country": "GB"
+            }
+        },
+        {
+            "tag": "hirebike",
+            "feed_url": "https://www.hirebikelincoln.co.uk/map-of-stations.html",
+            "meta": {
+                "latitude": 53.2145628869,
+                "longitude": -0.603208756027,
+                "city": "Lincoln",
+                "name": "Hirebike",
+                "country": "GB"
+            }
+        },
+        {
+            "tag": "sheffieldbycycle",
+            "feed_url": "https://www.sheffieldbicycle.co.uk/map-of-stations.html",
+            "meta": {
+                "latitude": 53.2145628869,
+                "longitude": -0.603208756027,
+                "city": "Sheffield",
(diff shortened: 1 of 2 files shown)


eskerda:
So the only thing different between bikeu and this is the regex to get stations?

I would say proposed solution would screw us more than unifying it with bikeu. Worst case we add more regexes. I can build that up parting from this PR, or if you want to work on it that's fine also.

Easiest, we would define a dict with different regexes like:

```
regexes = {
    'bikeu': r'<bikeu regex>',
    'hourbike': r'<hourbike regex>',
}
```

And then setting that key on the data file. We can even default it to bikeu.

As for defining the company, it can be set on the data file. It sucks to write it for each, but it's what we have right now.

As per naming, we do not really know what operates what, most probably bikeu and hourbike are using the same tech.

Other option would be moving the regex as a class attribute, and then creating another class subclassing bikeu. The data file would move to being one of these multiclass data files.

edumucelli:
@eskerda, no, BikeU takes the data from an iframe, all the code related to `id("MapData")/@src` does not exist on the Hourbike. It is not that the whole code is the same but the regex, it is that they do not share some of the fields, e.g., `'TKBike'`, `'TKStation'`. On the other hand, the two cities I've mentioned share the tech with BikeU.

eskerda:
Only one of the bikeu systems uses the iframe. The others do not use that code either

pybikes/data/hourbike.json
```diff
@@ -0,0 +1,50 @@
+{
+    "instances": [
+        {
+            "tag": "readybike",
+            "feed_url": "https://www.readybike.co.uk/map-of-stations.html",
```
eUgEntOptIc44, line 5: url does not exists anymore

pybikes/data/hourbike.json
```diff
@@ -0,0 +1,50 @@
...
+                "latitude": 53.2145628869,
+                "longitude": -0.603208756027,
+                "city": "Lincoln",
+                "name": "Hirebike",
+                "country": "GB"
+            }
+        },
+        {
+            "tag": "sheffieldbycycle",
+            "feed_url": "https://www.sheffieldbicycle.co.uk/map-of-stations.html",
```
eUgEntOptIc44, line 38: url also does not exists anymore

pybikes/data/hourbike.json
eUgEntOptIc44: also see: https://road.cc/content/news/218094-reading-bike-hire-schemes-future-chaos

eUgEntOptIc44 (changes requested):
many years later to of the urls don't exist anymore. guess @edumucelli has also moved on to new projects.

futuretap:
Both systems no longer exist.

PR #158, opened 2016-06-30
[smartbike] Add "Malmö by bike" from Malmö, SE

lbschenkel:
Malmö by bike <https://www.malmobybike.se/en> is a bike rental system in
Malmö (Sweden) operated by Clear Channel.

Since some stations do not contain 'address' and/or 'nearbyStations',
the class SmartBike was changed to cope with the missing fields and avoid
throwing errors.

P.S.: I don't have experience with Python, so apologies if my changes are not idiomatic.

### DIFF
--- a/pybikes/smartbike.py
+++ b/pybikes/smartbike.py
@@ -89,18 +89,19 @@ class SmartBikeStation(BikeShareStation):
             self.longitude = float(info['lon'])
             self.extra = {
                 'uid': int(info['id']),
-                'status': info['status'],
-                'address': info['address']
+                'status': info['status']
             }
-
+            if 'address' in info:
+                self.extra['address'] = info['address']
             if 'district' in info:
                 self.extra['districtCode'] = info['district']
             elif 'districtCode' in info:
                 self.extra['districtCode'] = info['districtCode']
 
-            if info['nearbyStations'] and info['nearbyStations'] != "0":
+            nearby_stations = info.get('nearbyStations')
+            if nearby_stations and nearby_stations != "0":
                 self.extra['NearbyStationList'] = map(
-                    int, info['nearbyStations'].split(',')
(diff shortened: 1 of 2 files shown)


pybikes/smartbike.py
```diff
@@ -90,15 +90,15 @@ def __init__(self, info):
             self.extra = {
                 'uid': int(info['id']),
                 'status': info['status'],
-                'address': info['address']
+                'address': info['address'] if 'address' in info else None
             }
```
eskerda: I would just declare address after like

```
if 'address' in info:
    self.extra['address'] = info['address']
```

pybikes/smartbike.py
```diff
@@ -90,15 +90,15 @@ def __init__(self, info):
...
 
             if 'district' in info:
                 self.extra['districtCode'] = info['district']
             elif 'districtCode' in info:
                 self.extra['districtCode'] = info['districtCode']
 
-            if info['nearbyStations'] and info['nearbyStations'] != "0":
+            if 'nearbyStations' in info and info['nearbyStations'] and info['nearbyStations'] != "0":
```
eskerda: If you want to check that it's there and that it's not empty, easiest is:

```
if info.get('nearbyStations') and info['nearbyStations'] != '0':
    ...
```

of course, this is also valid

```
nearby_stations = info.get('nearbyStations')
if nearby_stations and nearby_stations != '0':
    ...
```

lbschenkel:
Amended commit with your feedback.

eskerda:
Arh, sorry to be a pain, but lat / lng are off again, these should be good `55.6050, 13.0038`

lbschenkel:
Amended coordinates.

eskerda:
Looking good, thanks!

PR #159, opened 2016-06-30
[cyclocity] Add "Lundahoj" from Lund, SE

lbschenkel:


eskerda:
Looks good, thanks!

Only thing: I am a bit OCD with latitudes and longitudes on system instances. Ideally, we want the reference center of the city (and not the center of where stations are located). It's to ease API consumers into not running too many geocoding queries. For this case, this would be good I think:

Latitude: 55.7046601
Longitude: 13.1910073

lbschenkel:
For my PRs I used the coordinates listed in Wikipedia. I have amended the commit using the coordinates that you provided.

eskerda:
> For my PRs I used the coordinates listed in Wikipedia. I have amended the commit using the coordinates that you provided.

Yeah, I think wikipedia rounds location to minutes only.

eskerda:
@lbschenkel thanks again!

PR #160, opened 2016-06-30
[cyclocity] Gothenburg's bike sharing system is called "Styr & Ställ"

lbschenkel:


eskerda:
Great, thanks @lbschenkel !

PR #164, opened 2016-07-02
nextbike: add support for a different subdomain (used by mvgrad)

robbi5:
New version of #154 - now with a hostname field in the json, to use a whitelabeled nextbike like MVG Rad (Munich)

--
For the bike numbers approximation by getting `bike_types`, I'll open another PR.

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -166,11 +166,23 @@
             "city_uid": "138"
         },
         {
-            "domain": "mg",
+            "domain": "de",
             "tag": "nextbike-munchen",
             "meta": {
                 "latitude": 48.1392,
                 "city": "München",
+                "longitude": 11.5791,
+                "country": "DE"
+            },
+            "city_uid": "139"
+        },
+        {
+            "domain": "mg",
+            "tag": "mvgrad",
+            "meta": {
+                "latitude": 48.1392,
+                "city": "München",
+                "name": "MVG Rad",
                 "longitude": 11.5791,
                 "country": "DE"
             },
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -187,7 +187,8 @@
                 "country": "DE"
             },
             "city_uid": "268",
-            "hostname": "mvgrad.nextbike.net"
+            "hostname": "mvgrad.nextbike.net",
+            "bbox": [[11.72288, 48.248222], [11.36078, 48.06159]]
         },
         {
             "domain": "lv",
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -166,7 +166,7 @@
             "city_uid": "138"
         },
         {
-            "domain": "de",
+            "domain": "mg",
             "tag": "nextbike-munchen",
             "meta": {
                 "latitude": 48.1392,
@@ -174,7 +174,8 @@
                 "longitude": 11.5791,
                 "country": "DE"
             },
-            "city_uid": "139"
+            "city_uid": "268",
+            "hostname": "mvgrad.nextbike.net"
         },
         {
             "domain": "lv",
--- a/pybikes/nextbike.py
+++ b/pybikes/nextbike.py
@@ -33,7 +33,7 @@ class Nextbike(BikeShareSystem):
         self.url = BASE_URL.format(hostname=hostname, domain=domain)
         self.uid = city_uid
         self.bbox = None
-        if bbox is not None:
+        if bbox:
             self.bbox = box(bbox[0][0], bbox[0][1], bbox[1][0], bbox[1][1])
 
     def update(self, scraper=None):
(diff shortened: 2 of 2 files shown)


eskerda:
What's the main difference between the "münchen" feed at https://nextbike.net/maps/nextbike-live.xml?domains=de and the one at http://mvgrad.nextbike.net/maps/nextbike-live.xml?domains=mg

You mention one is a whitelabel instance of a nextbike. But what about the previous one, does it still exist, is it a ghost system? I see no direct correlation between the two but in case both co-exist, we shall do something.

robbi5:
Oh, I've just looked into the nextbike app on android - Munich is still there. Their official website says:

> Außerhalb der Münchner City – unter anderem in Planegg, Germering und Krailling – stehen weiterhin ca. 70 Fahrräder an mehreren nextbike-Stationen zur Verfügung.

> _Outside the city of Munich - including Planegg, Germering and Krailling - there are still about 70 bikes at several nextbike stations._

So it seems the center of munich is powered by MVG Rad (the whitelabeled nextbike) - and some stations outside keep the nextbike brand.

I think we should keep both. Should I add MVG Rad as a seperate entry to the json? :)

eskerda:
Checked the stations (I should do that more often). For the network we currently call `nextbike-munchen` these are the stations. Only one actually in there, and the others scattered around.

image

This is for the system you are adding. Looks good in Munich, 

image

But sadly, it also contains one station (or bike?) on Leipzig and on a train station somewhere else on the border...

image

image


Now I am slightly confused (as I usually am with nextbike and call a bike). What should we do with our current `nextbike-munchen`? Also, we should filter out these two ghost stations on the `mvgrad` feed.

It's not the first time we filter out invalid stations from feeds. We usually resort to some dumb method, like checking for invalid lat / lng or present things on the name string, or also something a bit more elegant like filtering stations outside the city / region bounding box [pybikes/youbike.py](https://github.com/eskerda/pybikes/blob/master/pybikes/youbike.py).

Ideally, if these systems are interoperable, we would merge `nextbike-munchen`and `mvgrad` together. That would involve some refactoring on the nextbike base too.

robbi5:
I think we should keep `nextbike-munchen` seperate. From a customers perspective these are two different systems - _MVG Rad_ is only bookable with the _MVG more_ app, _nextbike-munchen_ is only bookable with the official nextbike app.

For `mvgrad` I'll add a filter for the stations outside of munich.

Fun fact: The station in Leipzig is called "nextbike Ralfs Schreibtisch Leipzig" - it seems to be an debugging entry of one of the nextbike developers.

eskerda:
Sounds good to me.

> Fun fact: The station in Leipzig is called "nextbike Ralfs Schreibtisch Leipzig" - it seems to be an debugging entry of one of the nextbike developers.

What about the other one, named "Alcatraz"? Seems pretty high on the mountains :)

pybikes/nextbike.py
```diff
@@ -26,29 +28,38 @@ class Nextbike(BikeShareSystem):
...
     }
 
-    def __init__(self, tag, meta, domain, city_uid, hostname = 'nextbike.net'):
+    def __init__(self, tag, meta, domain, city_uid, hostname = 'nextbike.net', bbox = None):
         super(Nextbike, self).__init__(tag, meta)
         self.url = BASE_URL.format(hostname=hostname, domain=domain)
         self.uid = city_uid
+        self.bbox = None
+        if bbox is not None:
+            self.bbox = box(bbox[0][0], bbox[0][1], bbox[1][0], bbox[1][1])
```
eskerda: on most cases, unless you want a strict checking `if bbox is not None` == `if bbox`

pybikes/nextbike.py
```diff
@@ -26,29 +28,38 @@ class Nextbike(BikeShareSystem):
...
+            if 'bike' in place.attrib:
+                if place.attrib['bikes'] == "1" and place.attrib['bike'] == "1":
+                    continue
+            if self.bbox is not None:
+                lat = float(place.attrib['lat'])
+                lng = float(place.attrib['lng'])
+                coord = Point(lng, lat)
+                if not self.bbox.contains(coord):
+                    continue
+            yield place
```
eskerda: :+1:

pybikes/data/nextbike.json
```diff
@@ -177,6 +177,20 @@
...
+            "meta": {
+                "latitude": 48.1392,
+                "city": "München",
+                "name": "MVG Rad",
+                "longitude": 11.5791,
+                "country": "DE"
+            },
+            "city_uid": "268",
+            "hostname": "mvgrad.nextbike.net",
+            "bbox": [[11.72288, 48.248222], [11.36078, 48.06159]]
```
eskerda: :+1:

eskerda:
@robbi5 is it complete?

robbi5:
Yep :) thanks for the help.

eskerda:
Awesome work, thanks @robbi5!

PR #166, opened 2016-07-07
[bonopark] Added Madrid's BiciMAD network

bcaller:
The bike network for Madrid, Spain.

### DIFF
--- a/pybikes/bonopark.py
+++ b/pybikes/bonopark.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016, Ben Caller <bcaller@gmail.com>
+# Distributed under the LGPL license, see LICENSE.txt
+
+import json
+import hashlib
+
+from .base import BikeShareSystem, BikeShareStation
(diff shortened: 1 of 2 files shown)


pybikes/bonopark.py
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    def __init__(self, tag, meta, url):
+        super(Bonopark, self).__init__(tag, meta)
+        self.feed_url = url
+
+    def update(self, scraper=None):
+        if scraper is None:
+            scraper = utils.PyBikesScraper()
+        scraper.headers['Content-Type'] = 'application/json'
```
eskerda: Can we set this user agent  `Apache-HttpClient/UNAVAILABLE (java 1.4)`? Seems to be the one the app is using.

bcaller: Yes that would be the correct user agent, though for now it works without.

bcaller:
I have added the user agent.

eskerda:
Looks good, thanks

Issue #163, opened 2016-06-30
Feature request: scrape "Bycyklen" from Copenhagen, DK


lbschenkel:
The stations are listed here:
https://bycyklen.dk/en/find-a-bycykel/

It's possible to get a JSON structure with all stations by doing a GET:
https://bycyklen.dk/umbraco/Surface/BookingSurface/GetStationsList

Unfortunately the list above does not include the number of available bikes and I didn't find a way yet to scrape that. I'll comment here if I find out more.

I don't know Python so I'm afraid that implementing a new scraper from scratch is beyond my current skills.

eskerda:
Seems like availability can be found at:

```
curl 'https://bycyklen.dk/en/my-account/my-reservations/book-cykel-via-website/'  --data 'lat=55.673582&lng=12.564984&page=0' -H 'X-Requested-With: XMLHttpRequest'
```

And from there, iterating over pages. It might not be trivial to correlate that to the feed. Either by closest distance to the marker (each availability contains a static map), or by similarity on name.

eskerda:
I also see their "android app" is just the same website adapted to the layout, so no luck on there for reverse engineering anything.

eskerda:
> I don't know Python so I'm afraid that implementing a new scraper from scratch is beyond my current skills.

It's true this spider might get a bit tricky. However do not overestimate the task, in the end all spiders are the same.

Either you can give it a try and I will review it, or I can do it and you review it so next time you know how you would do it?

lbschenkel:
I can actually try implementing it, it'll be a good excuse to learn Python — however I believe that it'll take a while since I'm not going to have a lot of free time in the foreseeable future.

In case you prefer to do it, I don't mind at all — I am more interested that this data becomes available as soon as possible than in who actually implements it.

eskerda:
> In case you prefer to do it, I don't mind at all — I am more interested that this data becomes available as soon as possible than in who actually implements it.

Sure. It's also not available at http://data.kk.dk/ so adding this might make them rethink and provide a good feed.

Will leave the issue open and whoever takes it first does it

Issue #124, opened 2016-03-25
Add Oslo Bysykkel bike sharing network


tuxayo:
It doesn't seem to be supported and there is already a Python package that access it https://pypi.python.org/pypi/oslobysykkel/

Has the support of this network been considered before?

abadabanades:
Hi, I have been investigating a bit, The package you mention is completely outdated, since the API has changed.
I found an endpoint with the stations in their new website: https://oslobysykkel.no/api/internal/stations
```json
{"stations":[{"title":"Arkitekt Rivertz Plass",
              "subtitle":"i Uelands gate",
              "ready":true,
              "center": { "latitude": 59.934934,
                           "longitude": 10.749475}
             },
             {"title": "...",
              ...
```
However, the available bikes and free spots seem missing; instead the "ready" field appears: I guess it states whether an station has enough bikes available (who knows where is the threshold). Any idea where to find it? It seems it was available in the previous version

eskerda:
: /

Their mobile app does not even seem to work without a registered telephone number

:-1:

eskerda:
They used to have an spec available at http://smartbikeportal.clearchannel.no/public/mobapp/maq.asmx

eskerda:
Seems `Snart Klart` most probably means `planned station`

eskerda:
Ok, found this: http://oslo.map.webservice.sharebike.com:8888/

eskerda:
This is even more interesting

http://oslo.map.webservice.sharebike.com:8888/json/MapService/LiveStationData?APIKey=A97C2382-BA61-432F-9518-060DE2C208C6&SystemID=cityoslo

eskerda:
Not sure how down the rabbit hole we need to go

eskerda:
Jackpot: https://oslobysykkel.no/api/v1/stations/availability

tuxayo:
Great find! (I understand it's feasible?)

eskerda:
Whilst this is technically feasible, this is a private endpoint. Emailed them about it, assumedly they told me it's going to be opened soon.

willosof:
They said "within the week" to me over a month ago. Anyways. In the meanwhile - i found the same stuff you guys found (using the app and https MITM attack) and made a simple module for node.js:
https://www.npmjs.com/package/oslobysykkel

eskerda:
> They said "within the week" to me over a month ago. 

[...]

PR #167, opened 2016-07-08
[gobike] Added Bycyklen, Bysykkelen & Nu-Connect

bcaller:
For #163

### DIFF
--- a/pybikes/gobike.py
+++ b/pybikes/gobike.py
@@ -3,13 +3,13 @@
 # Distributed under the LGPL license, see LICENSE.txt
 
 import json
+from urlparse import urljoin
 
 from lxml import html
 from lxml.cssselect import CSSSelector
 
-from .base import BikeShareSystem, BikeShareStation
 from . import utils
-from contrib import TSTCache
+from .base import BikeShareSystem, BikeShareStation
 
 # The number of free slots is unavailable
 # However, you can lock the bike next to any full station
@@ -26,17 +26,19 @@ class GoBike(BikeShareSystem):
         'company': 'Gobike A/S'
     }
 
-    def __init__(self, tag, meta, url, availability_path):
+    def __init__(self, tag, meta, hostname, availability_path):
         super(GoBike, self).__init__(tag, meta)
-        self.url = url
-        self.availability_path = availability_path
+        self.availability_url = urljoin(hostname, availability_path)
+        self.stations_url = urljoin(hostname, STATION_LIST_PATH)
 
     def update(self, scraper = None):
         scraper = scraper or utils.PyBikesScraper()
 
         # Request station names and locations from JSON file
-        stations = json.loads(scraper.request(self.url + STATION_LIST_PATH))["List"]
-        stations_by_id = {str(s["UnifiedId"]): GoBikeStation(s) for s in stations}
+        stations = json.loads(scraper.request(self.stations_url))["List"]
+        stations_by_id = {
+            str(s["UnifiedId"]): GoBikeStation(s) for s in stations
+        }
 
         # Request number of bikes, at 10 stations per page
         for page in self._get_all_pages(scraper, len(stations)):
@@ -53,7 +55,7 @@ class GoBike(BikeShareSystem):
                 'lng': self.meta['longitude'],
                 'page': p
             }
-            yield scraper.request(self.url + self.availability_path, 'POST', data=data)
+            yield scraper.request(self.availability_url, 'POST', data=data)
 
     @staticmethod
     def _parse_page(body):
--- a/pybikes/gobike.py
+++ b/pybikes/gobike.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2015, Ben Caller <bcaller@gmail.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import json
+
+from lxml import html
+from lxml.cssselect import CSSSelector
+
+from .base import BikeShareSystem, BikeShareStation
+from . import utils
+from contrib import TSTCache
+
+__all__ = ['GoBike', 'GoBikeStation']
+
+PAGE_SIZE = 10
+STATION_LIST_PATH = '/umbraco/Surface/BookingSurface/GetStationsList'
+
+cache = TSTCache(delta=60)
+
+class GoBike(BikeShareSystem):
+    sync = True
+
+    meta = {
+        'system': 'GoBike',
+        'company': 'GoBike'
+    }
+
+    def __init__(self, tag, meta, url, availability_path):
+        super(GoBike, self).__init__(tag, meta)
+        self.url = url
+        self.availability_path = availability_path
+
+    def _get_availability_request(self, page, number_of_stations):
+        url = self.url + self.availability_path
+        data = {
+            'lat': self.meta['latitude'],
+            'lng': self.meta['longitude']
+        }
+        if number_of_stations <= PAGE_SIZE:
+            return url, 'POST', data
+
+        data['page'] = page
+        # The query string is not really necessary, but is added to help with the caching
+        return '{}?p={}'.format(url, page), 'POST', data
+
+    def update(self, scraper = None):
+        if scraper is None:
+            scraper = utils.PyBikesScraper(cache)
+
+        station_list = json.loads(scraper.request(self.url + STATION_LIST_PATH))["List"]
+        stations_by_id = dict((str(s["UnifiedId"]), GoBikeStation(s)) for s in station_list)
+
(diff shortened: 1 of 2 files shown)


bcaller:
#124

pybikes/gobike.py
```diff
@@ -0,0 +1,104 @@
...
+            if self.extra['zip']:
+                self.extra['address'] += ', ' + self.extra['zip']
+                if self.extra['city']:
+                    self.extra['address'] += ' ' + self.extra['city']
+            elif self.extra['city']:
+                self.extra['address'] += ', ' + self.extra['city']
+
+    def set_available_bikes(self, n):
+        if self.bikes:
+            raise Exception("Already")
```
bcaller: This check is unnecessary. Removing it...

pybikes/gobike.py
```diff
@@ -0,0 +1,102 @@
...
+from .base import BikeShareSystem, BikeShareStation
+from . import utils
+from contrib import TSTCache
+
+__all__ = ['GoBike', 'GoBikeStation']
+
+PAGE_SIZE = 10
+STATION_LIST_PATH = '/umbraco/Surface/BookingSurface/GetStationsList'
+
+cache = TSTCache(delta=60)
```
eskerda: The TSTCache is meant to be used on cases where two systems point to the same URL (for example, nextbike). Is this the case for this system?

pybikes/gobike.py
```diff
@@ -0,0 +1,102 @@
...
+            if self.extra['zip']:
+                self.extra['address'] += ', ' + self.extra['zip']
+                if self.extra['city']:
+                    self.extra['address'] += ' ' + self.extra['city']
+            elif self.extra['city']:
+                self.extra['address'] += ', ' + self.extra['city']
+
+    def set_available_bikes(self, n):
+        self.bikes = n
+        # self.free is unavailable, but with this system you can lock the bike next to a full station
```
eskerda: This helper function is not really necessary

pybikes/gobike.py
```diff
@@ -0,0 +1,102 @@
...
+            if self.extra['street_building_identifier']:
+                self.extra['address'] += ' ' + self.extra['street_building_identifier']  # Number comes after in DK
+            if self.extra['district']:
+                self.extra['address'] += ', ' + self.extra['district']
+            if self.extra['zip']:
+                self.extra['address'] += ', ' + self.extra['zip']
+                if self.extra['city']:
+                    self.extra['address'] += ' ' + self.extra['city']
+            elif self.extra['city']:
+                self.extra['address'] += ', ' + self.extra['city']
```
eskerda: For readability, could be translated to following (or similar)

```python
class GoBikeStation(BikeShareStation):
    def __init__(self, info):
        super(GoBikeStation, self).__init__()
        location = info['Location']
        self.name = info['Name']
        self.latitude = float(location['Latitude'])
        self.longitude = float(location['Longitude'])
        self.extra = {
            'uid': str(info['UnifiedId']),
            'status': int(info['Status']),
            'address': self.format_address(location),
        }

    def format_address(self, location):
        address_sections = [
            ['Street', 'StreetBuildingIdentifier'],
            ['DistrictName'],
            ['ZipCode', 'City'],
        ]
        address = []
        for section in address_sections:
            components = [location.get(k) for k in section]
            components = filter(None, components)
            part = ' '.join(components)
            address.append(part)
        address = filter(None, address)
        address = ', '.join(address)

        return address
```

pybikes/gobike.py
```diff
@@ -0,0 +1,102 @@
...
+        # Can only query 10 stations per request, perhaps we should do this async
+        for page in range(0, len(station_list)/PAGE_SIZE + (len(station_list) % PAGE_SIZE > 0)):
+            body = scraper.request(*self._get_availability_request(page, len(station_list)))
+            dom = html.fromstring(body)
+            for e in station_selector(dom):
+                uid = e.get('id').split('_')[1]
+                bikes = int(available_selector(e)[0].text)
+                stations_by_id[uid].set_available_bikes(bikes)
+
+        self.stations = stations_by_id.values()
```
eskerda: Code flow is a bit confusing, what about something like (untested)

```python
PAGE_SIZE = 10
STATION_LIST_PATH = '/umbraco/Surface/BookingSurface/GetStationsList'


class GoBike(BikeShareSystem):
    sync = True

    meta = {
        'system': 'GoBike',
        'company': 'GoBike'
    }

    def __init__(self, tag, meta, url, availability_path):
        super(GoBike, self).__init__(tag, meta)
        self.hostname = url
        self.availability_url = urljoin(self.hostname, availability_path)
        self.stations_url = urljoin(self.hostname, STATION_LIST_PATH)

    def update(self, scraper=None):
        scraper = scraper or utils.PyBikesScraper()

        stations = json.loads(scraper.request(self.stations_url))['List']
        stations_by_id = {
            str(s['UnifiedId']): GoBikeStation(s) for s in stations
        }

        for page in self.get_pages(scraper, len(stations)):
            for uid, bikes in self.parse_page(page):
                stations_by_id[uid].bikes = bikes

        self.stations = stations_by_id.values()

    def get_pages(self, scraper, n_stations):
        n_pages = (n_stations / PAGE_SIZE) + (n_stations % PAGE_SIZE > 0)
        for i in range(0, n_pages):
            data = {
                'lat': self.meta['latitude'],
                'lng': self.meta['longitude'],
                'page': i,
            }
            yield scraper.request(self.availability_url, method='POST',
                                  data=data)

    def parse_page(self, body):
        station_selector = CSSSelector('.span6 .station-basicinfo')
        available_selector = CSSSelector('.station-availablebikes')
        dom = html.fromstring(body)
        for e in station_selector(dom):
            uid = e.get('id').split('_')[1]
            bikes = int(available_selector(e)[0].text)
            yield (uid, bikes)
```

eskerda:
About doing it async, current library is not ready for this case (async systems update each station, so it's "easy" to map).

Something better, like a function exposing all async tasks with callbacks would do that possible, but won't happen without a major rewrite of pybikes.

pybikes/gobike.py
bcaller: Ah no. They have different domains.

pybikes/gobike.py
bcaller: Yes, this is nice.

bcaller:
Your suggested changes were great. What do you think about the new version?

pybikes/gobike.py
```diff
@@ -0,0 +1,98 @@
...
+# Distributed under the LGPL license, see LICENSE.txt
+
+import json
+
+from lxml import html
+from lxml.cssselect import CSSSelector
+
+from .base import BikeShareSystem, BikeShareStation
+from . import utils
+from contrib import TSTCache
```
eskerda: Unused import

pybikes/gobike.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, tag, meta, url, availability_path):
+        super(GoBike, self).__init__(tag, meta)
+        self.url = url
+        self.availability_path = availability_path
+
+    def update(self, scraper = None):
+        scraper = scraper or utils.PyBikesScraper()
+
+        # Request station names and locations from JSON file
+        stations = json.loads(scraper.request(self.url + STATION_LIST_PATH))["List"]
```
eskerda: urljoin is much nicer than appending strings

pybikes/gobike.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _get_all_pages(self, scraper, n_stations):
+        n_pages = n_stations/PAGE_SIZE + (n_stations % PAGE_SIZE > 0)
+        for p in range(0, n_pages):
+            data = {
+                'lat': self.meta['latitude'],
+                'lng': self.meta['longitude'],
+                'page': p
+            }
+            yield scraper.request(self.url + self.availability_path, 'POST', data=data)
```
eskerda: for instance, this append is done for each page, when it could have been constructed on __init__, since the URL is always the same

pybikes/gobike.py
```diff
@@ -0,0 +1,98 @@
...
+        ]
+        address = []
+        for section in address_sections:
+            components = filter(None, (location.get(k) for k in section))
+            part = ' '.join(components)
+            address.append(part)
+        address = filter(None, address)
+        address = ', '.join(address)
+
+        return address
```
eskerda: :+1:

eskerda:
Couple of comments, about creating urls by appending and an unused import.

bcaller:
Just wondering... will citybik.es return null or 0 for these stations with `free_bikes` set to `None`? I suppose I'll find out.

eskerda:
> Just wondering... will citybik.es return null or 0 for these stations with free_bikes set to None? I suppose I'll find out.

It will return null.

PR #169, opened 2016-07-08
nextbike: count bikes by using get_biketypes

robbi5:
_Extracted from #154_
The URL parameter `get_biketypes` adds another field to the resulting xml, `bike_types`. It looks like that these are [the different bike types](https://iframe.nextbike.net/4com/map/?t=btg) and the number available on this station.

Example:
```xml
<place ... bikes="5+" ... bike_types="{"15":10}"/>
```

We can use this to replace the approximate "5+" count with a more realistic number :)

### DIFF
--- a/pybikes/nextbike.py
+++ b/pybikes/nextbike.py
@@ -3,6 +3,7 @@
 # Distributed under the AGPL license, see LICENSE.txt
 
 import re
+import json
 from lxml import etree
 
 from shapely.geometry import Point, box
@@ -13,7 +14,7 @@ from pybikes.contrib import TSTCache
 
 __all__ = ['Nextbike', 'NextbikeStation']
 
-BASE_URL = 'https://{hostname}/maps/nextbike-live.xml?domains={domain}'
+BASE_URL = 'https://{hostname}/maps/nextbike-live.xml?domains={domain}&get_biketypes=1'
 CITY_QUERY = '/markers/country/city[@uid="{uid}"]/place'
 
 cache = TSTCache(delta=60)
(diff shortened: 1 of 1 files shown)


eskerda:
Mind if I join in? It seems nextbike code would love a bit of cleaning up. After exporting all nextbike stations into a csv and putting it on pandas. https://gist.github.com/eskerda/7de77874cc087c615be40a804685735c

Summary:

1. Filter bikes from stations by checking if bike is set and is '1'.
2. Do not try to regex anything on the name, get it raw
3. Get number when available, under extra.number.
4. Get uid under extra.uid.
5. On cases where there's bike_types available, trust these
6. else, trust bikes, and if bikes is 5+, set it as extra.bikes_approximate = True.
7. free_racks and bike_racks are always available together (it means, both set or not).
8. Trust free_racks as free
9. Trust bike_racks as extra.slots
10. When no rack info available, set slots to ?? (undecided)

I think I will branch from here, and do the proposed cleanup, if you are ok with it?

robbi5:
Yeah, looks great :)

---

From your notes:
> There are cases where there are more bikes than bike_racks. This means we must not rely on setting free_slots to bike_racks - bikes.

This seems to be expected behaviour —see this excerpt from their [FAQ (sorry, german only)](https://www.mvg.de/services/mobile-services/mvg-rad/mvg-rad-faq.html):
>Was mache ich, wenn die Station voll ist?
>Stellen Sie das Rad einfach direkt neben der Station ab. Ihnen werden die Freiminuten trotzdem gut geschrieben. Bitte achten Sie darauf, Fluchtwege nicht zu versperren.
>
>_What should I do if the station is full?_
>_Place the bike just next to the station. You'll still get the free minutes. Please be careful not to block escape routes._

eskerda:
Closed in favour of https://github.com/eskerda/pybikes/pull/171

PR #171, opened 2016-07-10
Fix nextbike code

eskerda:
Starting from https://github.com/eskerda/pybikes/pull/169 and under the assumptions on https://gist.github.com/eskerda/7de77874cc087c615be40a804685735c

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -916,7 +916,7 @@
             "city_uid": "283"
         },
         {
-            "domain": "at",
+            "domain": "ta",
             "tag": "nextbike-serfaus",
             "meta": {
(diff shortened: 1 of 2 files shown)


eskerda:
This cleans up nextbike code. For cases where no slots are available, it is set to `None`, which would be represented on the API as `null`. Usually, we marked such cases with `-1`, but that seems to bring on confusion.

Thoughts about `-1` vs `null` vs not setting it at all? Now it's the time to speak, pinging @edumucelli @robbi5 @gszathmari @BenSto @bparmentier @javnik36 @thesebas @felixhaass

eskerda:
@robbi5 thanks for the idea of using `bike_types`! all shall be good now

PR #174, opened 2016-07-15
List install dependencies

mvl22:
This commit adds the required install dependencies to the readme, giving an example set of commands (for Ubuntu, but will be understandable by sysadmins for other systems).


eskerda:
looks good, thanks!

Issue #175, opened 2016-07-15
[nextbike] Warwick 


bcaller:
There is a next bike service in Warwick, UK. Not sure how big it is. We should maybe include it.

PR #176, opened 2016-07-16
[nextbike] add university of warwick (located in Coventry not Warwick)

bcaller:
#175

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -857,6 +857,18 @@
             "city_uid": "267"
         },
         {
+            "domain": "uk",
+            "tag": "nextbike-uni-warwick",
+            "meta": {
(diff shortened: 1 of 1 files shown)


eskerda:
As far as I see, this system seems to have a name (if not, it will default to `Nextbike`).

https://nextbike.net/maps/nextbike-live.xml?domains=uk reports "University of Warwick", but that's hardly a name.

The known name seems to be 'UniCycles' or even 'UniCycles @ Warwick'.

- http://www2.warwick.ac.uk/about/environment/transport/cycling/uni-cycles/
- http://www.nextbike.co.uk/en/university%20of%20warwick/

bcaller:
Yes I copied from one which didn't have a name field. Whoops.

eskerda:
:+1:  Looks good, thanks!

PR #181, opened 2016-07-24
[Smoove] Removing Reflex, service is over

edumucelli:
Service was supressed September 2015. You can read more info [here](http://france3-regions.francetvinfo.fr/bourgogne/saone-et-loire/chalon-sur-saone/saone-et-loire-pourquoi-chalon-range-ses-velos-reflex-au-garage-760686.html) (in French)


eskerda:
:+1:

PR #182, opened 2016-08-03
Call superclass' __init__

flebel:
`CleanapStation` is missing a call to its `BikeShareStation`'s `__init__` to set default values.


eskerda:
Thanks @flebel!

PR #184, opened 2016-08-17
Nextbike: fix bike_types 5+ breaking a lot of systems

edumucelli:
`bike_types` also brought the `5+` ugliness to a lot of systems. This PR fixes it.


eskerda:
Seems they changed it, too bad. This works for now as a hotfix, but I do not like supporting two approximate methods.

PR #187, opened 2016-08-25
Baksi implementation

anastmag:
I almost implemented baksi system. It needs some minor modifications (override BikeShareStation.\__str__ function in order to properly print Turkish letters, set sync and unifeed variable) but it works (i think that i should set sync=True but i can't find out where unifeed variable is used). If i find some time tomorrow i'll fix them.

### DIFF
--- a/pybikes/baksi.py
+++ b/pybikes/baksi.py
@@ -54,13 +54,13 @@ class Baksi(BikeShareSystem):
 class BaksiStation(BikeShareStation):
     def __init__(self, data):
         super(BaksiStation, self).__init__()
-        self.station_id=data[0]
         self.name=data[1]
-        self.status=data[2]
         self.bikes=data[3]
         self.free=data[4]
         self.latitude=float(data[5])
         self.longitude=float(data[6])
         self.extra={
+                'uid': data[0]
+                'status': data[2]
                 'slots': int(data[3]+data[4]),
         }
--- a/pybikes/baksi.py
+++ b/pybikes/baksi.py
@@ -2,7 +2,7 @@
 # Copyright (C) 2010-2012, eskerda <eskerda@gmail.com>
 # Distributed under the AGPL license, see LICENSE.txt
 
-from .base import import BikeShareSystem, BikeShareStation
+from .base import BikeShareSystem, BikeShareStation
 from . import utils
 from contrib import TSTCache
 
@@ -11,4 +11,22 @@ __all__ = ['Baksi', 'BaksiStation']
 cache = TSTCache(delta=60)
 
 class Baksi(BikeShareSystem):
+    #sync =
+    #unifeed =
+    meta = {
+        'system': 'Baksi',
+        'company': ['Baksi Bike Sharing System']
+    }
 
+    def __init__(self, tag, meta, city_uid, feed_url):
+        super(Baksi, self).__init__(tag, meta)
+        self.feed_url=feed_url
+
+    def update(self, scraper=None):
+        if scraper is None:
+            scraper = utils.PyBikesScraper()
+
+
+class BaksiStation(BikeShareStation):
+    def __init__(self, info):
+        pass
--- a/pybikes/baksi.py
+++ b/pybikes/baksi.py
@@ -10,13 +10,11 @@ from contrib import TSTCache
 
 __all__ = ['Baksi', 'BaksiStation']
 
-# Atomic bomb rgx
-# Tomorrow
-# Use unicode to transform to utf-8
-ID_NAME_RGX="[0-9]*\-[a-zA-Z0-9\s]*"
-STATUS_RGX=""
-DOCKS_BIKES_RGX=""
-LAT_LNG_RGX=""
+ID_NAME_RGX="([0-9]+)\-([\w\s.()-]+)\'"
+STATUS_RGX="Durum\ [&nbsp;]+\ (\w+)"
+DOCKS_RGX="Park[&nbsp;]+([0-9]+)"
+BIKES_RGX="Bisiklet[&nbsp;]+([0-9]+)"
+LAT_LNG_RGX="([\s0-9.]+)\',\ \'([\s0-9.]+)"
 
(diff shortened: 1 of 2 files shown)


pybikes/baksi.py
```diff
@@ -0,0 +1,72 @@
...
+DOCKS_RGX="Park[&nbsp;]+([0-9]+)"
+BIKES_RGX="Bisiklet[&nbsp;]+([0-9]+)"
+LAT_LNG_RGX="([\s0-9.]+)\',\ \'([\s0-9.]+)"
+
+USERAGENT = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/31.0.1650.63 Chrome/31.0.1650.63 Safari/537.36"
+
+cache = TSTCache(delta=60)
+
+class Baksi(BikeShareSystem):
+    #sync = ???
```
eskerda: All systems are sync by default. It just means there are no intermediate steps to get full information (like, let's say, you get stations and then, for each station, you need to do a request too).

This is a crude explanation: https://eskerda.com/categorizing-citybikes-data-feeds/

pybikes/baksi.py
```diff
@@ -0,0 +1,72 @@
...
+
+ID_NAME_RGX="([0-9]+)\-([\w\s.()-]+)\'"
+STATUS_RGX="Durum\ [&nbsp;]+\ (\w+)"
+DOCKS_RGX="Park[&nbsp;]+([0-9]+)"
+BIKES_RGX="Bisiklet[&nbsp;]+([0-9]+)"
+LAT_LNG_RGX="([\s0-9.]+)\',\ \'([\s0-9.]+)"
+
+USERAGENT = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/31.0.1650.63 Chrome/31.0.1650.63 Safari/537.36"
+
+cache = TSTCache(delta=60)
```
eskerda: TSTCache is only used for systems where it makes sense to save the previous result (imagine a system where all subsystems request the same URL, we do not want to hit their server 200 times if response is the same). On this case, I think it's not needed, as each network has a different endpoint

pybikes/baksi.py
```diff
@@ -0,0 +1,72 @@
...
+        """Refine output
+        """
+        station_id, name=zip(*id_name)
+        status=["Active" if out=="Aktif" else "Inactive" for out in status]
+        docks=[int(i) for i in docks]
+        bikes=[int(i) for i in bikes]
+        latitude, longitude=zip(*geopoints)
+
+        for data in zip(station_id, name, status, docks, bikes, latitude, longitude):
+            self.stations.append(BaksiStation(data))
```
eskerda: This will keep adding stations on the same `Baksi` instance, if `update()` is called more than one time over the same obj. So it either must be initialized with an empty list, or a new list must be used and then assigned to `self.stations`

pybikes/baksi.py
```diff
@@ -0,0 +1,72 @@
...
+
+__all__ = ['Baksi', 'BaksiStation']
+
+ID_NAME_RGX="([0-9]+)\-([\w\s.()-]+)\'"
+STATUS_RGX="Durum\ [&nbsp;]+\ (\w+)"
+DOCKS_RGX="Park[&nbsp;]+([0-9]+)"
+BIKES_RGX="Bisiklet[&nbsp;]+([0-9]+)"
+LAT_LNG_RGX="([\s0-9.]+)\',\ \'([\s0-9.]+)"
+
+USERAGENT = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/31.0.1650.63 Chrome/31.0.1650.63 Safari/537.36"
```
eskerda: Is the USERAGENT really necessary, as in, does it work without it? We only override configured user agent in case system does not work (not a prospective ban).

anastmag:
So far i have to solve the remaining two issues.

1. > This will keep adding stations on the same Baksi instance, if update() is called more than one time over the same obj. So it either must be initialized with an empty list, or a new list must be used and then assigned to self.stations

2. Override BikeShareSystem.\__str__ or find a way to properly print Turkish characters.

I'll take care of them tomorrow.

pybikes/baksi.py
anastmag: deleted lines 19,39

pybikes/baksi.py
anastmag: deleted lines 9,21

pybikes/baksi.py
anastmag: removed commented lines

pybikes/baksi.py
anastmag: replaced lines 59,60 with
`self.stations=map(BaksiStation, zip(station_id, name, status, docks, bikes, latitude, longitude)`

I can also pass a dictionary to BaksiStation as an argument but i should modify BaksiStation.\__init__. Let me know if you want me to do that.

anastmag:
I can't find out what's going on with the proper printing of Turkish characters. Any ideas ?

bcaller:
The page uses the Turkish encoding ISO-8859-9

pybikes/baksi.py
```diff
@@ -0,0 +1,66 @@
...
+        bikes=[int(i) for i in bikes]
+        latitude, longitude=zip(*geopoints)
+
+        self.stations=map(BaksiStation, zip(station_id, name, status, docks, bikes, latitude, longitude))
+
+
+class BaksiStation(BikeShareStation):
+    def __init__(self, data):
+        super(BaksiStation, self).__init__()
+        self.station_id=data[0]
```
eskerda: There's no station_id base field. It's on extra under `extra['uid']`

pybikes/baksi.py
```diff
@@ -0,0 +1,66 @@
...
+        self.name=data[1]
+        self.status=data[2]
```
eskerda: This also goes to extra

eskerda:
:+1: thanks @anastmag !!

PR #189, opened 2016-09-04
[velobike-ru] Remove station ID from address extra field

bcaller:
Station name is e.g. `466 - Продольный пр-д (у главного входа ВДНХ)`, but the address field should not include the station ID at the start.

### DIFF
--- a/pybikes/velobike_ru.py
+++ b/pybikes/velobike_ru.py
@@ -46,7 +46,7 @@ class VelobikeRU(BikeShareSystem):
             extra = {
                 'uid': item['Id'],
                 'slots': int(item['TotalPlaces']),
-                'address': item['Address'],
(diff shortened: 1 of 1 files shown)


eskerda:
@bcaller WDYT about using a regex substitution like
```
>>> text = "466 - Продольный пр-д (у главного входа ВДНХ)"
>>> print re.sub(r'^\d+\s*-\s*', '', text)
Продольный пр-д (у главного входа ВДНХ)
```

bcaller:
Yeah that works. But why is it better?

eskerda:
Because it explicitly does what you want it to do: "remove a number that appears at the start of a string and is followed by an hyphen".

What does `'Foo - Bar'.split("- ", 1)[1]` do?

eskerda:
:+1:

Issue #165, opened 2016-07-06
Make "company" fields in "networks" elements always an array


brescia123:
Currently the field `company` inside `networks` elements is sometimes a single string and sometimes an array. 
I think it is better and more standardized to make it **always an array** of one or more items.

eskerda:
```
$ git grep "company[\"|\']: [\"|\']"
README.md:    'company': 'PBSC',
pybikes/adcb.py:        'company': 'Cyacle Bicycle Rental LLC'
pybikes/bicicard.py:        'company': 'ITCL'
pybikes/bicincitta.py:        'company': 'Comunicare S.r.l.'
pybikes/bikeshare_ie.py:        'company': 'The National Transport Authority'
pybikes/bikeu.py:        'company': 'Bike U Sp. z o.o.'
pybikes/bixi.py:        'company': 'PBSC'
pybikes/callabike.py:        'company': 'DB Rent GmbH'
pybikes/cleanap.py:        'company': 'CleaNap'
pybikes/clujbike.py:        'company': 'Municipiul Cluj-Napoca'
pybikes/cyclocity.py:        'company': 'JCDecaux',
pybikes/cyclocity.py:        'company': 'JCDecaux'
pybikes/cyclopolis.py:        'company': 'Cyclopolis Systems'
pybikes/decobike.py:        'company': 'DecoBike LLC'
pybikes/domoblue.py:        'company': 'Domoblue'
pybikes/ecobici_ba.py:        'company': 'Buenos Aires Ciudad',
pybikes/emovity.py:        'company': 'ICNITA S.L.'
pybikes/encicla.py:        'company': 'Sistema de Bicicletas Públicas del Valle de Aburrá'
pybikes/gewista_citybike.py:        'company': 'Gewista Werbegesellschaft m.b.H'
pybikes/keolis.py:        'company': 'Keolis'
pybikes/keolis.py:        'company': 'Keolis'
pybikes/labici.py:        'company': 'Labici Bicicletas Públicas SL',
pybikes/mvgmeinrad.py:        'company': 'Mainzer Verkehrsgesellschaft mbH (MVG)'
pybikes/nextbike.py:        'company': 'Nextbike GmbH'
pybikes/punpunbikeshare.py:        'company': 'BTS Group Holdings',
pybikes/smartbike.py:        'company': 'ClearChannel'
pybikes/smoove.py:        'company': 'Smoove'
[...]

PR #193, opened 2016-09-09
Make company field always an array

brescia123:
As described in #165 I think that `company` field should be always an array to make easier for clients to consume the JSON.

### DIFF
--- a/pybikes/bixi.py
+++ b/pybikes/bixi.py
@@ -24,7 +24,7 @@ class BixiSystem(BikeShareSystem):
 
     meta = {
         'system': 'Bixi',
-        'company': 'PBSC'
+        'company': ['PBSC']
     }
 
     def __init__(self, tag, feed_url, meta, format):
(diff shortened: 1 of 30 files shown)


pybikes/bixi.py
```diff
@@ -168,22 +168,22 @@ def from_json(data):
...
+            "installed": "true",
+            "locked": "false",
+            "installDate": "1313724600000",
+            "removalDate": {  },
+            "temporary": "false",
+            "public": "true",
+            "nbBikes": "15",
+            "nbEmptyDocks": "8",
+            "latestUpdateTime": "1375592453128"
         }
```
eskerda: This part of the commit is not necessary, unless you do not mind appeariong on the `git blame` for it

eskerda:
Looks good. We should probably add some tests so new systems conform this rule, will do that on another PR

pybikes/bixi.py
brescia123: Of course, that was a mistake! 😅

eskerda:
Looks good, thanks!

PR #194, opened 2016-09-09
get_biketypes is now provided by default, default API host: api.nextb…

j0kan:
…ike.net


robbi5:
Great 👍 
But it's currently not the default on `mvgrad.nextbike.net`, so requests for munich still require `get_biketypes=1`

j0kan:
You're right so rather wait another 2-3 weeks until mvgrad has the latest software, too.

Altonss:
Is this PR still relevant 6 years later, or could it be closed?

j0kan:
Still relevant - should be merged.

PR #195, opened 2016-09-12
Gbfs parser implementation

anastmag:
Alright, i implemented the bare bones of the gbfs parser. 

In the GbfsStation class (derived from BikeShareStation) i included only the variables latitude, longitude, free, bikes, station_id, name and status. I also added a file named gbfs.json in pybikes/data directory with all bike share systems that at the moment support gbfs spec (you can find a list of them right here [https://github.com/NABSA/gbfs/blob/master/systems.csv](url)). 

In this way, it's pretty straightforward to add values to BikeShareStation.extra:
1. Cut the certain system from data/gbfs.json and paste it to a new json file.
2. Override GbfsStation class and add the needed fields to BikeShareStation.extra by writing a new pybikes/'system'.py file

Note: Some tags in gbfs.json conflict with other systems (for example both gbfs.json and bcycle.json include 'indego' tag). I think that it's meaningful to remove duplicates.

Waiting for your feedback.


eskerda:
> In the GbfsStation class (derived from BikeShareStation) i included only the variables latitude, longitude, free, bikes, station_id, name and status. I also added a file named gbfs.json in pybikes/data directory with all bike share systems that at the moment support gbfs spec (you can find a list of them right here https://github.com/NABSA/gbfs/blob/master/systems.csv).

:+1: 

> In this way, it's pretty straightforward to add values to BikeShareStation.extra:
> 1. Cut the certain system from data/gbfs.json and paste it to a new json file.
> 2. Override GbfsStation class and add the needed fields to BikeShareStation.extra by writing a new pybikes/'system'.py file

Does this mean that not all gbfs feeds support the same fields?

> Note: Some tags in gbfs.json conflict with other systems (for example both gbfs.json and bcycle.json include 'indego' tag). I think that it's meaningful to remove duplicates.

Sure, any feed that supports gbfs can be removed from other providers

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+from .base import BikeShareSystem, BikeShareStation
+from . import utils
+
+__all__=['Gbfs', 'GbfsStation']
+
+class Gbfs(BikeShareSystem):
+
+    def __init__(self, tag, meta, feed_url):
+        super(Gbfs, self).__init__(tag, meta)
+        self.feed_url=feed_url
```
eskerda: Can we make feed_url available in meta here so we can expose it later on the API?

Something like

```
meta['gbfs_href'] = feed_url
```

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+            feeds[feed['name']]=feed['url']
+
+        '''Station Information and Station Status data retrieval
+        '''
+        station_information=json.loads(scraper.request(feeds['station_information']))
+        station_status=json.loads(scraper.request(feeds['station_status']))
+
+        for info, status in zip(station_information['data']['stations'],\
+        station_status['data']['stations']):
+            self.stations.append(GbfsStation(info, status))
```
eskerda: This will keep appending stations upon

```
foo = Gbfs(...)
foo.update()
foo.update()
```

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+        Example status variable:
+        {u'last_reported': 1473627463, u'num_bikes_disabled': 3, u'num_docks_available': 16,
+        u'is_renting': 1, u'station_id': u'7000', u'num_docks_disabled': 0, u'is_installed':
+        1, u'num_bikes_available': 12, u'is_returning': 1}
+
+        So let's extract the dataaa
+
+        Note: Maybe it's valuable to check every time the station_ids to see if
+        they match.
+        """
```
eskerda: If it's a possibility that order does not match, both lists can be grouped together by `station_id` before zipping.

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+        super(GbfsStation, self).__init__()
+
+        self.station_id=[info['station_id'] if
+                info['station_id']==status['station_id'] else 'Wrong']
```
eskerda: There's no `station_id` in `BikeShareStation`. We save this under: `extra['uid']`

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+        self.name=str(info.get('name'))
+        self.status=['Active' if status['is_renting'] and
+            status['is_installed'] else 'Inactive']
+        self.bikes=status.get('num_bikes_available')
+        self.free=status.get('num_docks_available')
+        self.latitude=float(info.get('lat'))
+        self.longitude=float(info.get('lon'))
```
eskerda: Code is going to fail anyway with a TypeError for `float(None)`, so IMHO there's no need to use `.get`

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        super(GbfsStation, self).__init__()
+
+        self.station_id=[info['station_id'] if
+                info['station_id']==status['station_id'] else 'Wrong']
+        self.name=str(info.get('name'))
+        self.status=['Active' if status['is_renting'] and
+            status['is_installed'] else 'Inactive']
+        self.bikes=status.get('num_bikes_available')
+        self.free=status.get('num_docks_available')
```
eskerda: Let's convert these to `int`, just in case some feeds give it out as str

pybikes/gbfs.py
```diff
@@ -0,0 +1,71 @@
...
+        Note: Maybe it's valuable to check every time the station_ids to see if
+        they match.
+        """
+        super(GbfsStation, self).__init__()
+
+        self.station_id=[info['station_id'] if
+                info['station_id']==status['station_id'] else 'Wrong']
+        self.name=str(info.get('name'))
+        self.status=['Active' if status['is_renting'] and
+            status['is_installed'] else 'Inactive']
```
eskerda: We have no `status` field on `BikeShareStation`. Some networks we support have status, and we have not a final decision on a good way to store status, so what about:

```
self.extra['open'] = all([status['is_renting'], status['is_installed']])
```

pybikes/gbfs.py
eskerda: last reported might be useful to know how trustworthy the info is (even if we have updated it)
```
self.extra['last_update'] = info['last_reported']
```

pybikes/gbfs.py
anastmag: I forgot it again.

anastmag:
>>  In this way, it's pretty straightforward to add values to BikeShareStation.extra:
    1. Cut the certain system from data/gbfs.json and paste it to a new json file.
    2. Override GbfsStation class and add the needed fields to BikeShareStation.extra by writing a new pybikes/'system'.py file

>Does this mean that not all gbfs feeds support the same fields? 

Because this is just a basic implementation of gbfs (you can add to 'extra' much more items about alerts, pricing etc.) if someone wants to add another value he is free to do it following these steps. Maybe it's my misunderstanding about 'self.extra' dictionary. What really should be stored in 'extra' regarding the great amount of data provided for every station by gbfs ? And what about this variable generally ? In both my pull requests (baksi and gbfs) i inspected already implemented systems in order to find out what values should be stored there.

Should i concentrate all possible station data provided from all gbfs_href's .json files to variable 'info' even if they are never used ? So far, i use only data from 'station_information.json' and 'station_status.json' because i tried to provide exactly the same values as every other system.

Sorry for any inconveniences.

eskerda:
> Should i concentrate all possible station data provided from all gbfs_href's .json files to variable 'info' even if they are never used ? So far, i use only data from 'station_information.json' and 'station_status.json' because i tried to provide exactly the same values as every other system.
> Sorry for any inconveniences.

Worry not. Most of these fields are not necessary ATM. I think it will be a good idea to write an script that generates instances (on the instance `data/` format) and there we can add more meta info to the systems, but that will come later.

On extra usually goes whatever is not part of the base station required fields (name, lat, lng, bikes, free) but that might be useful. They are on a separate part for people using the lib or the API to know that these are not consistent among networks. In case we manage to get something on extra consistent (uid), a field might be moved from extra to the base item.

pybikes/data/gbfs.json
```diff
@@ -0,0 +1,523 @@
...
+                "city": "University of Virginia",
+                "name": "Ubike",
+                "country": "US"
+            },
+            "feed_url": "https://ubike.virginia.edu/opendata/gbfs.json"
+        }
+    ],
+    "system": "gbfs",
+    "class": "Gbfs"
+}
```
eskerda: Noticed some networks are changing tags. That's ATM a bad idea, since system relies on networks being consistent (ie: someone consuming `sanantonio` feed will find out that it does not exist anymore and that it is now `bcycle_sanantonio`). There is no system in there currently to properly redirect or record network id changes except done manually on the API side. I know this is a boring task, so I will send a PR to this one.

Some notes about style:
- Add whitespace around operators (`i=10` -> `i = 10`)
- Block comments always start with `# `
- Do not use `\` for multiline expressions

For instance

```
        #Station Information and Station Status data retrieval
        station_information=json.loads(scraper.request(feeds['station_information']))\
            ['data']['stations']
```

Might become
```
        # Station Information and Station Status data retrieval
        station_information = json.loads(
            scraper.request(feeds['station_information'])
        )['data']['stations']
```

More info at https://www.python.org/dev/peps/pep-0008/ (no need to follow it strictly). It's always a good idea to have a python linter on the editor. I use `python-mode` plugin for vim.

anastmag: At least now i think that i understand the extra variable.
Thank your for mentioning pep-8, i've been pretty much careless. Don't bother with the tags, i'll keep the old tags instead of the new ones. Just give me some time.

anastmag: Maybe there's a way that we can map one system to many tags if we want to add also the 'bcycle_' prefix to all bcycle systems that are implemented at the moment. But before that we should define a naming convention.

eskerda:
@anastmag I fixed some tedious problems on current revision, check it out at https://github.com/anastmag/pybikes/pull/2

eskerda:
All looks good to me. Thanks a lot @anastmag!

PR #196, opened 2016-09-13
[gobike] Support XML endpoint

eskerda:


eskerda:
@bcaller asking you since you initially added this system.

Is it a big deal if we do not get address on this new type of feed?

PR #198, opened 2016-09-18
added bike sharing cities for http://www.mimuovoinbici.it/

MatteoBatti:
added bike sharing datas for the following italian cities:  …
Folì
Cesena
Imola
Modena
San Giovanni in Persiceto
San Lazzaro di Avena
Scandiano

### DIFF
--- a/pybikes/data/bicincitta.json
+++ b/pybikes/data/bicincitta.json
@@ -335,6 +335,83 @@
                 },
                 {
                     "meta": {
+                        "latitude": 44.157801,
+                        "city": "Cesena",
+                        "name": "MiMuovoCe",
+                        "longitude": 12.239802,
+                        "country": "IT"
+                    },
+                    "endpoint": "http://www.mimuovoinbici.it/frmLeStazioni.aspx?ID=120",
+                    "tag": "mimuovoce"
+                },
+                {
+                    "meta": {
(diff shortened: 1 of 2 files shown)


eskerda:
@MatteoBatti sorry, didn't see the updated PR.  Where do the names of these systems come from, ie: MiMuovoCe, MiMuovoFo, ...

Assuming the photos of the bike share systems, it seems the name is simply "Mi Muovo in bici"?

image

MatteoBatti:
I used these names Because I saw the feed for the city of Ferrara named "MiMuovoFe", i clould named them all "Mi Muovo in bici"

eskerda:
Yeah.. that came from here https://github.com/eskerda/pybikes/pull/111 I assumed that was the name of the system on Ferrara, but I think the system is just named "Mi Muovo in bici".

Do you think we should name all systems contained in http://www.mimuovoinbici.it/frmLeStazioni.aspx as "Mi Muovo in bici" ?

MatteoBatti:
I checked with a friend of mine that live in Ferrara, the system is called "Mi Muovo in bici". We can change the name and use "Mi Muovo in bici" for all the systems under http://www.mimuovoinbici.it/frmLeStazioni.aspx

eskerda:
All added in another PR

PR #207, opened 2016-10-22
[Easybike] Fix and cleanup

edumucelli:
EasyBike in Greece changed its URL, not formally, but as their API in the /cities.php does not work anymore, I have found a new URL. Some cities are not available in this new feed, in those cases the city map is gone, as the case of [nafpaktos](http://nafpaktos.easybike.gr/el/map) and [keratsini-drapetsona](http://keratsini-drapetsona.easybike.gr/el/map).

Besides, I have found Athens feed in there.

### DIFF
--- a/pybikes/data/easybike.json
+++ b/pybikes/data/easybike.json
@@ -64,18 +64,7 @@
                 "longitude": 24.4122164,
                 "country": "GR"
             },
-            "city_uid": "kavala_new"
-        },
-        {
-            "tag": "easybike-drapetsona",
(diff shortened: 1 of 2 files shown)


eskerda:
@edumucelli thanks for the heads up. Added some changes and merged in https://github.com/eskerda/pybikes/pull/218

Can I ask you to be more careful adding the lat / lngs? The ones you used on this PR point nowhere https://goo.gl/maps/pgThfoosk9D2

PR #209, opened 2016-10-23
Add support for bicicorunha system

dvilchez:


eskerda:
Looks good. Unfortunately, we can't really use `suds` at the moment, since it's required that all systems use `PyBikesScraper` interface for requesting any url.

Also, yesterday we found out that the endpoint `/swm/WSMobile.asmx` can also be used for http://www.biciferrolterra.com/swm/WSMobile.asmx

dvilchez:
why can't we use suds? SOAP (https://en.wikipedia.org/wiki/SOAP) is actually a pretty complex protocol and I don't think you can overcome all this intrinsic complexity with a HTTP request and some X path...

anyway, feel free to change the code to suit your requisites.

eskerda:
> why can't we use suds? SOAP (https://en.wikipedia.org/wiki/SOAP) is actually a pretty complex protocol and I don't think you can overcome all this intrinsic complexity with a HTTP request and some X path...
>

I completely agree with that. Plus, having integration tests is quite cool too.

The problem is not on using or not suds, but a flaw on pybikes design. Because suds does not use the shared scraper class, any modification to the scraper (ie, configuring a proxy) passed along to the update method won't be respected by suds without an interface between them. This is completely doable, but lowers the benefits of using this library.

About SOAP, it's complex but because we only depend on one call it's actually pretty easy (maybe ugly). You can check current implementation at [pybikes/nextgal.py](https://github.com/eskerda/pybikes/blob/master/pybikes/nextgal.py)

eskerda:
On Oct 24, 2016 11:52 AM, "dvilchez" <notifications@github.com> wrote:
>
> why can't we use suds? SOAP (https://en.wikipedia.org/wiki/SOAP) is
actually a pretty complex protocol and I don't think you can overcome all
this intrinsic complexity with a HTTP request and some X path...
>

I completely agree with that. Plus, having integration tests is quite cool
too.

The problem is not on using or not suds, but a flaw on pybikes design.
Because suds does not use the shared scraper class, any modification to the
scraper (ie, configuring a proxy) passed along to the update method won't
be respected by suds without an interface between them. This is completely
doable, but lowers the benefits of using this library.

About SOAP, it's complex but because we only depend on one call it's
actually pretty easy (maybe ugly). Check current implementation at
https://github.com/eskerda/pybikes/blob/master/pybikes/nextgal.py

dvilchez:
sorry @eskerda I was off the last two weeks, thanks for fixing the code.

PR #211, opened 2016-10-24
Add missing minus sign to Tampa longitude

BenSto:
Otherwise the coordinates point to a place in Nepal instead of Tampa, FL. Feel free to have a look and merge into master if you agree that Tampa, FL is not in Nepal ;-) . :bike:  :bike:  :grinning:


eskerda:
I do agree, thanks @BenSto!

PR #213, opened 2016-10-25
Add CompartiBike

edumucelli:
Adding CompartiBike, 3 cities in Brazil.

PR #214, opened 2016-10-25
[Samba] Sorocaba is over within Mobilicidade system

edumucelli:
Sorocaba's system is not on Mobilicidade's anymore, it was improved and went to CompartiBike, refer to #213.


eskerda:
Looks good, thanks!

PR #223, opened 2016-11-11
[Velobike_ru] Fix feed_url

edumucelli:
Feed URL has changed, updating it.


eskerda:
Both URLs seem to work ATM. Let's close this and in case it fails again later we can revisit.

Thanks @edumucelli

PR #228, opened 2016-12-17
[Cleanap] project is suspended

edumucelli:
Project's [App](https://play.google.com/store/apps/details?id=it.lumilab.movinap) description says that "THE RESEARCH PROJECT IS ACTUALLY SUSPENDED" and the site's map does not show anymore.


eskerda:
Good to know, thanks @edumucelli!

PR #230, opened 2017-01-24
[bilbonbizi] Added Bilbao's Bilbon Bizi network

artberri:
Bona nit ;)

I've added the bicycle-sharing system of Bilbao called 'Bilbon Bizi':
http://www.bilbao.eus/cs/Satellite/agenda21/Servicio-Prestamo-de-Bicicletas-Bilbon-Bizi/es/100078677/Contenido_Detalle

The source is the Open Data portal of Bilbao's Council: 
https://www.bilbao.eus/opendata/es/catalogo/dato-puntos-recogida-bicicletas-en-prestamo

Let me know if I need to fix something in order to merge the PR.

Congratulations for this great project!!!

### DIFF
--- a/pybikes/bilbonbizi.py
+++ b/pybikes/bilbonbizi.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2017, Alberto Varela <alberto@berriart.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+from lxml import etree
+
+from pybikes.base import BikeShareSystem, BikeShareStation
+from pybikes.utils import PyBikesScraper
+
+__all__ = ['BilbonBizi']
+
+class BilbonBizi(BikeShareSystem):
+    sync = True
+
+    meta = {
+        'system': 'BilbonBizi',
+        'company': ['Ayuntamiento de Bilbao'],
+        'license': {
+            'name': 'Creative Commons Reconocimiento 3.0 España',
+            'url': 'https://creativecommons.org/licenses/by/3.0/es/'
+        },
+        'source': 'https://www.bilbao.eus/opendata/es/catalogo/dato-puntos-recogida-bicicletas-en-prestamo'
+    }
+
+    def __init__(self, tag, meta, feed_url):
+        super(BilbonBizi, self).__init__(tag, meta)
+        self.feed_url = feed_url
+
+    def update(self, scraper=None):
+        if scraper is None:
+            scraper = PyBikesScraper()
+        raw = scraper.request(self.feed_url, 'GET', None, None, True)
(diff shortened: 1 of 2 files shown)


pybikes/data/bilbonbizi.json
```diff
@@ -0,0 +1,17 @@
...
+        {
+            "meta":{
+                "city": "Bilbao",
+                "name": "Bilbon Bizi",
+                "latitude": 43.2627707,
+                "longitude": -2.9371822,
+                "country": "ES"
+            },
+            "feed_url": "http://www.bilbao.eus/WebServicesBilbao/WSBilbao?s=ODPRESBICI&u=OPENDATA&p0=A&p1=A",
+            "tag": "bilbonbizi"
```
eskerda: Being the name `Bilbon Bizi` I think `bilbon-bizi` as a tag sounds better

pybikes/bilbonbizi.py
```diff
@@ -0,0 +1,45 @@
...
+
+__all__ = ['BilbonBizi']
+
+class BilbonBizi(BikeShareSystem):
+    sync = True
+
+    meta = {
+        'system': 'BilbonBizi',
+        'company': ['Ayuntamiento de Bilbao']
+    }
```
eskerda: Since this API has a license (not that usual) can you add a `license` field similar to https://github.com/eskerda/pybikes/blob/master/pybikes/cyclocity.py#L35?

artberri:
Hi @eskerda , requested changes already done! I hope everything is fine.

One extra question: 

I'm thinking about adding also the bike sharing service of San Sebastian called dBizi. But after investigating about it, I think that the only viable way to do it is by scraping their website. I don't see any public API (neither private, it seems that they print the information as inline javascript in their webpage when showing the station map).

So, is it valid to crawl the website or I shouldn't add them until they have an API?

eskerda:
Looks good, thanks!

> I'm thinking about adding also the bike sharing service of San Sebastian called dBizi. But after investigating about it, I think that the only viable way to do it is by scraping their website. I don't see any public API (neither private, it seems that they print the information as inline javascript in their webpage when showing the station map).
> So, is it valid to crawl the website or I shouldn't add them until they have an API?

Actually, there's this endpoint somewhere http://www.donostia.eus/info/ciudadano/camaras_trafico.nsf/dameEstaciones , but since it displays different info than the website and its not documented I think it's better to just scrape the website. We do so in many cases, even if it's a nasty regex for getting inline js. In this case it's fairly easy, since its a raw JSON.

artberri:
Thanks for merging!

About Donostia, yeah, I know about that URL also but it is always displaying 'free slots: 0' for every station, and as you said it seems easy to crawl :+1:

PR #236, opened 2017-04-01
GBFS: Remove coast bikeshare

edumucelli:
It is gone from GBFS.

### DIFF
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -270,21 +270,6 @@
             "feed_url": "https://gbfs.citibikenyc.com/gbfs/gbfs.json"
         },
         {
-            "tag": "coast-bike-share",
-            "meta": {
-                "latitude": 27.9627193250593,
-                "longitude": -82.438056400783,
-                "city": "Tampa, FL",
-                "name": "Coast Bike Share",
-                "country": "US",
-                "company": [
-                    "CycleHop, LLC",
-                    "Social Bicycles Inc."
-                ]
-            },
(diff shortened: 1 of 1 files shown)


eskerda:
Two systems seem to protrude from coast bike share:
- http://coastbikeshare.com/stpete/
- http://coastbikeshare.com/tampa/map/

edumucelli:
We should start moving some of the missing SocialBycicles systems on GBFS to its dedicated parser. I will bootstrap something.

Altonss:
It seams like the Coast Bikeshare in Tampa is now operated by hopr, which has a gbfs feed here: https://gbfs.hopr.city/api/gbfs/8/

Altonss:
Well if you look in the details, this feed also contains classical docked bikes (on stations) :)
So personally I would not close this issue. Maybe add the docked bikes for now, before we support floating bikes in #469

eskerda:
It does, but they do not have latitude or longitude. Anyway, I modified the feed https://github.com/eskerda/pybikes/commit/42bb9ac5aa63d7d246ac1e4dde20b2eb131ee9c6 just so we can keep track of it without having this old issue opened

Altonss:
> It does, but they do not have latitude or longitude.

The are available here: https://gbfs.hopr.city/api/gbfs/8/station_information

eskerda:
Then that's a bug on their systems and also on our gbfs implementation. That happens when one of the two entries do not have a field and get overwritten on the dictionary update. Look at gbfs.py on this line https://github.com/eskerda/pybikes/blob/27dc6ea1ef23121238366468698dba07dd1ba927/pybikes/gbfs.py#L92

Maybe you want to take a look

PR #237, opened 2017-04-01
GBFS: Remove Toronto

edumucelli:
Looks like it does not exist anymore


flebel:
It appears to me that the service still exists but the feed URL changed: https://bikesharetoronto.com/system-map/

eskerda:
So much for the GBFS format.. Seems a lot of systems do not really care

eskerda:
Good news, https://github.com/eskerda/pybikes/pull/243 updates toronto GBFS feed

Thanks!

PR #239, opened 2017-04-01
Nextbike: Remove mvgrad

edumucelli:
Its stations [URL](https://mvgrad.nextbike.net/maps/nextbike-live.xml?domains=mg&get_biketypes=1) is not there anymore.

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -177,20 +177,6 @@
             "city_uid": "139"
         },
         {
-            "domain": "mg",
-            "tag": "mvgrad",
-            "meta": {
-                "latitude": 48.1392,
-                "city": "München",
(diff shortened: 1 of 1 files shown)


eskerda:
Meh, I would say they added auth to it, so services inside mvg can keep consuming the feed...

Found these:
- From their website: https://carsharing.mvg-mobil.de/json/stations.php
- From their app: https://multimobil-core.mvg.de/service/v6/networkState/networkState?MVG_RAD=0
- nextbike produces 9 stations on another feed: http://nextbike.net/maps/nextbike-live.xml?domain=de (city uid 139)

Not sure what we should do. Probably deleting them out of existence is the best idea.

On a happier line, I found this json which might come useful:
- https://api.nextbike.net/maps/nextbike-live.json?list_cities=1

PR #240, opened 2017-04-03
[nextbike] Networks changes in Warsaw

adamdanielczyk:
Removed no longer existing 'Bemowo Bike' which was replaced by 'Stacje Sponsorskie Nextbike Veturilo'.

Also, two other minor changes: 
- added .idea/ to .gitignore
- fixed reference-style links in readme


eskerda:
Looks good. Thanks!

PR #242, opened 2017-04-04
GBFS: Pronto Cycle Share ceased operations on March 31, 2017

edumucelli:
As shown in the news and already displayed in [their](https://www.prontocycleshare.com/) site.


eskerda:
:+1:

PR #246, opened 2017-04-12
Changed Call-a-bike

autumnus:
Changed Call-a-bike to actual website setting

### DIFF
--- a/pybikes/data/callabike.json
+++ b/pybikes/data/callabike.json
@@ -1,149 +1,486 @@
 {
     "instances": [
         {
-            "tag": "callabike-darmstadt",
+            "tag": "callabike-aschaffenburg",
             "meta": {
-                "latitude": 49.8707135,
-                "city": "Darmstadt",
-                "name": "Call-A-Bike Darmstadt",
-                "longitude": 8.6527299,
+                "latitude": 49.436467,
+                "city": "Aschaffenburg",
+                "name": "Call-A-Bike Aschaffenburg",
+                "longitude": 7.768645,
                 "country": "DE"
-            },
-            "city_id": 19
+            }
         },
         {
-            "tag": "konrad",
+            "tag": "callabike-baden-baden",
             "meta": {
-                "latitude": 51.3148861,
-                "city": "Kassel",
-                "name": "Konrad",
-                "longitude": 9.4606148,
+                "latitude": 48.790514,
+                "city": "Baden Baden",
+                "name": "Call-A-Bike Baden Baden",
+                "longitude": 8.191302,
                 "country": "DE"
-            },
-            "city_id": 34100
+            }
+        },
+        {
+            "tag": "callabike-bamberg",
+            "meta": {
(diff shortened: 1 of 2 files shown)


pybikes/data/callabike.json
```diff
@@ -3,10 +3,10 @@
         {
             "tag": "callabike-aschaffenburg",
             "meta": {
-                "latitude": 49.436467,
```
piotrowski: These lat,lon values are the center of the "Geschäftsgebiet" for each city?

autumnus: To be honest: I don't know. I took a look at the data in request that is made, when a city is chosen. These coordinates are the values from the request. e.g.
{"method":"Map.listBikes","params":[{"lat":"52.519444","long":"13.406667","maxItems":100,"radius":100000}],"id":1492592072925} for Berlin

autumnus: What I just realized is, that the "id" seems to be a timestamp. So maybe there has to be a code change. I will have look at that.

piotrowski: Ok at "request that is made". I ask because then it is quite possible that not all bikes are returned because of the radius, right? Also limited to 100 definitely means not all bikes are returned, at least in big cities.

autumnus: Good point. I made a few tries: The lat/long seems to be the initial center of the search. This changes if you type an address into the search mask. So you are right: you will not get all stations with one request (DAMN!).
I think I will not have so much time to make changes at the moment. But feel free to play around with my code. Hopefully it is at least a good basis.

piotrowski: It sure is a 100% improvement to the non-working code right now :)
But really fixing this might be harder than expected...

eskerda: That's exactly the problem. There are three options:

1. As @piotrowski commented, an incomplete version being better than none
2. Splice the space for each network into a grid of coordinates, involving multiple calls. There's an API call to get the boundaries of each system (see https://github.com/eskerda/pybikes/issues/238#issuecomment-292649216). Together with some utility functions, the grid can be optimized to reduce the number of calls.
3. Wait until they provide a better version of their API. Contact them and complain. There's no reason for this data to not be provided cleanly.

swoker: @piotrowski  @eskerda Hi, I'm from the IT of Deutsche Bahn Connect who's the operator of Call a Bike.  
Feel free to use our Open Data API diretly! That is probably way easier than parsing our website :)    
http://data.deutschebahn.com/dataset/flinkster-api

Have fun and feel free to contact me for further questions!

eskerda: @swoker that's great news. Thank you!

eskerda:
Closing this PR, as flinkster API support will include new code (and possibly deleting old call a bike code).

eskerda:
@autumnus thanks for giving it a try, discovery of flinkster API was a success too!

PR #251, opened 2017-05-29
Add Tel-O-Fun

aronsky:
Added a module for Tel-O-Fun, Tel Aviv's bike sharing system.

### DIFF
--- a/pybikes/fsm.py
+++ b/pybikes/fsm.py
@@ -36,10 +36,13 @@ class FSMSystem(BikeShareSystem):
         for placemark in dom.xpath('//kml:Placemark', namespaces=ns):
             name = placemark.findtext('kml:name', namespaces=ns)
             info = placemark.findtext('kml:description', namespaces=ns)
-            station_uid, bikes, free = utils.re.findall(r'\w+\:\s*(\d+)', info)
-            latitude, longitude = placemark.findtext( \
-                                  'kml:Point/kml:coordinates', \
-                                    namespaces=ns).split(',')
+            station_uid, bikes, free = map(int,
+                utils.re.findall(r'\w+\:\s*(\d+)', info)
+            )
+            latitude, longitude = placemark.findtext(
+                'kml:Point/kml:coordinates',
+                namespaces=ns
+            ).split(',')
             latitude = float(latitude)
             longitude = float(longitude)
             extra = {
--- a/pybikes/fsm.py
+++ b/pybikes/fsm.py
@@ -9,7 +9,7 @@ from pybikes import utils
 
 class FSMSystem(BikeShareSystem):
     sync = True
-    unifeed = True
+    unifeed = False
 
     meta = {
         'system': 'fsm',
--- a/pybikes/fsm.py
+++ b/pybikes/fsm.py
@@ -7,17 +7,17 @@ from pybikes.base import BikeShareSystem, BikeShareStation
 from pybikes import utils
 
 
-class TeloFun(BikeShareSystem):
+class FSMSystem(BikeShareSystem):
     sync = True
     unifeed = True
 
     meta = {
-        'system': 'telofun',
+        'system': 'fsm',
         'company': ['FSM Ground Services Ltd.']
     }
 
     def __init__(self, tag, meta, feed_url):
-        super(TeloFun, self).__init__(tag, meta)
+        super(FSMSystem, self).__init__(tag, meta)
 
         self.feed_url = feed_url
 
--- a/pybikes/telofun.py
+++ b/pybikes/telofun.py
@@ -37,18 +37,14 @@ class TeloFun(BikeShareSystem):
                    .nodeValue
             info = placemark.getElementsByTagName('description')[0] \
                    .childNodes[0].nodeValue
-            station_code, bikes, free = [
-                    int(line.strip().split(':')[-1].strip()) \
-                    for line in info.strip().split('<br/>') if len(line) > 0
-                    ]
+            station_uid, bikes, free = utils.re.findall(r'\w+\:\s*(\d+)', info)
             latitude, longitude = placemark.getElementsByTagName('Point')[0] \
                                   .childNodes[0].childNodes[0].nodeValue \
                                   .split(',')
             latitude = float(latitude)
             longitude = float(longitude)
             extra = {
-                'code': station_code,
-                'slots': bikes + free
+                'uid': station_uid,
             }
 
             station = BikeShareStation(name, latitude, longitude, bikes, free,
--- a/pybikes/telofun.py
+++ b/pybikes/telofun.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright (C) 2017, aronsky <aronsky@gmail.com>
 
-from xml.dom import minidom
+from lxml import etree
 
 from pybikes.base import BikeShareSystem, BikeShareStation
 from pybikes import utils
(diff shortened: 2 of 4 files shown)


pybikes/telofun.py
```diff
@@ -0,0 +1,56 @@
...
+        dom = minidom.parseString(data.encode('utf-8', errors='ignore'))
+        stations = self.get_stations(dom)
+        self.stations = list(stations)
+
+    def get_stations(self, dom):
+        for placemark in dom.getElementsByTagName('Placemark'):
+            name = placemark.getElementsByTagName('name')[0].childNodes[0] \
+                   .nodeValue
+            info = placemark.getElementsByTagName('description')[0] \
+                   .childNodes[0].nodeValue
```
eskerda: Whilst using minidom is absolutely ok, sometimes it can result on a involved solution. See same implementation using lxml:

```
from lxml import etree
...
ns = {'kml': 'http://www.opengis.net/kml/2.2'}
dom = etree.fromstring(data.encode('utf-8'))
placemarks = dom.xpath('//Placemark', namespaces=ns)
for placemark in placemarks:
    name = placemark.findtext('kml:name', namespaces=ns)
    ...
```

Since minidom is part of the standard library the decision of using lxml or minidom is up to you, just though it was a good FYI

pybikes/telofun.py
```diff
@@ -0,0 +1,56 @@
...
+                    int(line.strip().split(':')[-1].strip()) \
+                    for line in info.strip().split('<br/>') if len(line) > 0
+                    ]
+            latitude, longitude = placemark.getElementsByTagName('Point')[0] \
+                                  .childNodes[0].childNodes[0].nodeValue \
+                                  .split(',')
+            latitude = float(latitude)
+            longitude = float(longitude)
+            extra = {
+                'code': station_code,
```
eskerda: If this is a unique value identifying stations, we add these as `extra.uid` instead of `extra.code`

pybikes/telofun.py
```diff
@@ -0,0 +1,56 @@
...
+                'slots': bikes + free
```
eskerda: Unless we get the total slots directly from the canonical source we usually don't add them, since these could not account of broken slots or broken bikes.

pybikes/telofun.py
```diff
@@ -0,0 +1,56 @@
...
+    def get_stations(self, dom):
+        for placemark in dom.getElementsByTagName('Placemark'):
+            name = placemark.getElementsByTagName('name')[0].childNodes[0] \
+                   .nodeValue
+            info = placemark.getElementsByTagName('description')[0] \
+                   .childNodes[0].nodeValue
+            station_code, bikes, free = [
+                    int(line.strip().split(':')[-1].strip()) \
+                    for line in info.strip().split('<br/>') if len(line) > 0
+                    ]
```
eskerda: This is somewhat involved too. Regular expressions aren't my cup of tea either, but they will break as easy as this, so for readability I would:
```
station_code, bikes, free = re.findall(r'\w+\:\s*(\d+)', info)
```

More complex queries could be written, such as querying exactly for `Station Code:` or `Available Bikes`, but for now, if this does not break, it's completely alright.

aronsky:
Hi, @eskerda, thanks for the comments! I'll make the appropriate changes ASAP.

pybikes/telofun.py
aronsky: I know minidom is not the best solution, but I chose it to avoid adding new dependencies for the project.  If you think that adding a dependency in this case is a better solution rather than staying with built-in libraries, I'll do it.

pybikes/telofun.py
aronsky: Will fix.

pybikes/telofun.py
aronsky: It is a unique value. Will change it.

pybikes/telofun.py
aronsky: Understood. No problem.

eskerda:
We already depend on lxml, so :+1: for using it

aronsky:
Okay, finished fixing up everything. I also changed the name of the system from 'telofun' to 'fsm', since I now more clearly understand the difference between a system and an instance.

Cheers!

pybikes/fsm.py
```diff
@@ -0,0 +1,51 @@
...
+        dom = etree.fromstring(data.encode('utf-8'))
+        stations = self.get_stations(dom)
+        self.stations = list(stations)
+
+    def get_stations(self, dom):
+        ns = {'kml': 'http://www.opengis.net/kml/2.2'}
+        for placemark in dom.xpath('//kml:Placemark', namespaces=ns):
+            name = placemark.findtext('kml:name', namespaces=ns)
+            info = placemark.findtext('kml:description', namespaces=ns)
+            station_uid, bikes, free = utils.re.findall(r'\w+\:\s*(\d+)', info)
```
eskerda: Got this:

```
Testing fsm
=============================
Testing u'Tel-O-Fun', u"Tel Aviv-Yaffo, Giv'atayim, Ramat-Gan"
Tel-O-Fun has 206 stations
[▚] testing 1F
======================================================================
FAIL: test_fsm (__main__.TestSystems)
Traceback (most recent call last):
  File "tests/unittest_pybikes.py", line 367, in test_schema
    self._test_systems(schema)
  File "tests/unittest_pybikes.py", line 27, in _test_systems
    self._test_system(instance['tag'], key)
  File "tests/unittest_pybikes.py", line 47, in _test_system
    self._test_station_ok(p_sys, p_sys.stations[i])
  File "tests/unittest_pybikes.py", line 70, in _test_station_ok
    self.assertIsInstance(station.bikes, int)
AssertionError: '4' is not an instance of <type 'int'>
```

Thought we were casting to int on constructor, but seems not. bikes and free must be `int`s.

pybikes/fsm.py
```diff
@@ -0,0 +1,51 @@
...
+
+from lxml import etree
+
+from pybikes.base import BikeShareSystem, BikeShareStation
+from pybikes import utils
+
+
+class FSMSystem(BikeShareSystem):
+    sync = True
+    unifeed = True
```
eskerda: This escaped me before, `unifeed` attribute is only needed for systems that with one call produce more than one instance and is just meant for a reference on code using `pybikes` to know that it should use a cache. Must either be deleted or set to `False`.

My bad, this should be documented.

pybikes/fsm.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def get_stations(self, dom):
+        ns = {'kml': 'http://www.opengis.net/kml/2.2'}
+        for placemark in dom.xpath('//kml:Placemark', namespaces=ns):
+            name = placemark.findtext('kml:name', namespaces=ns)
+            info = placemark.findtext('kml:description', namespaces=ns)
+            station_uid, bikes, free = utils.re.findall(r'\w+\:\s*(\d+)', info)
+            latitude, longitude = placemark.findtext( \
+                                  'kml:Point/kml:coordinates', \
+                                    namespaces=ns).split(',')
```
eskerda: Not really important, but since there's the issue with `bikes` and `free`, this style can be fixed too. No need to use backslashes here for line continuation:

```
latitude, longitude = placemark.findtext(
    'kml:Point/kml:coordinates',
    namespaces=ns
).split(',')
```
or even
```
latitude, longitude = placemark.findtext("""
    kml:Point/kml:coordinates
""", namespaces=ns).split(',')
```

pybikes/fsm.py
aronsky: Done. I converted the ```uid``` to ```int``` as well, since it is an integer.

aronsky:
Done.

eskerda:
Great, @aronsky. Thanks for your contribution!

aronsky:
Cheers!

PR #261, opened 2017-06-13
Add warning about correct Python version

gregopet:
This should prevent programmers not familiar with the Python ecosystem from losing half an hour before they can contribute their bike sharing systems to the project.


eskerda:
For the time being this will suffice. What I should actually do is publish pybikes on pypi and get some tags.

Thanks @gregopet!

PR #268, opened 2017-07-17
Urban Infrastructure Partner (bergen bysykkel and oslo bysykkel)

olavxxx:


pybikes/data/bysykkel.json
```diff
@@ -0,0 +1,34 @@
...
+        "latitude":59.913869,
+        "country":"NO",
+        "name":"Oslo bysykkel",
+        "longitude":10.752245,
+        "city":"Oslo",
+        "city_code":"OSL"
+      },
+      "feed_url":"https://oslobysykkel.no/api/v1/stations",
+      "feed_details_url":"https://oslobysykkel.no/api/v1/stations/availability",
+      "feed_token":"b14e14eec39cccdf0140e728db2012b9"
```
eskerda: Such keys, unless this is somehow a public key and not a developer requested one, should not be packaged with the library.

Systems that require a key can be instanced with a key parameter (see velib example on the README). This key will be passed as `key` parameter on instantiation.

pybikes/bysykkel.py
```diff
@@ -0,0 +1,86 @@
...
+    meta = {
+        'system': 'BySykkel',
+        'company': ['Urban Infrastructure Partner']
+    }
+
+    def __init__(self, tag, meta, feed_url, feed_details_url, feed_token):
+        super(BySykkel, self).__init__(tag, meta)
+        self.feed_url = feed_url
+        self.feed_details_url = feed_details_url
+        self.feed_token = feed_token
```
eskerda: If a system is using a key, it should set its class attribute as `authed = True`. See [cyclocity.py](https://github.com/eskerda/pybikes/blob/master/pybikes/cyclocity.py) as an example

pybikes/bysykkel.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def update(self, scraper=None):
+        if scraper is None:
+            scraper = utils.PyBikesScraper()
+            scraper.setUserAgent("pybikes - github.com/eskerda/pybikes/tree/master/pybikes")
```
eskerda: User agent should not be specified (ie: should be the default) unless it's necessary for the spider to actually work. Unless, of course, it's specified on the terms for https://developer.oslobysykkel.no/api

pybikes/bysykkel.py
```diff
@@ -0,0 +1,86 @@
...
+
+        for item in stations_data['stations']:
+            station = BikeShareStation()
+
+            station.name = item['title']
+
+            station.longitude = float(item['center']['longitude'])
+            station.latitude  = float(item['center']['latitude'])
+
+            details_filtered = [obj for obj in details_data['stations'] if(obj['id'] == item['id'])]
```
eskerda: I know performance is not really key for a spider, but this loop seems wasteful (on the order of n²).

By indexing both lists of stations by their station id, we can match them linearly. See [gbfs.py implementation](https://github.com/eskerda/pybikes/blob/master/pybikes/gbfs.py#L46-L52) for reference.

pybikes/data/bysykkel.json
```diff
@@ -0,0 +1,34 @@
...
+        "city_code":"BGO"
+      },
+      "feed_url":"https://bergenbysykkel.no/api/internal/stations",
+      "feed_details_url":"https://bergenbysykkel.no/api/internal/stations/availability",
+      "feed_token":"missing developer option :-/"
+    }
+  ],
+  "system":"bysykkel",
+  "class":"BySykkel"
+}
```
eskerda: Even though it was not an informed decision at the time, all (or most?) json data feeds are indented to 4 spaces.

olavxxx:
Cleaned up code line

pybikes/data/bysykkel.json
olavxxx: Removed the feed-token and implemented as requested.

pybikes/data/bysykkel.json
olavxxx: Implemented 4 spaces indentation as requested

pybikes/bysykkel.py
olavxxx: Implemented

pybikes/bysykkel.py
olavxxx: Removed the user agent

pybikes/bysykkel.py
olavxxx: Removed the inner loop and refactored the code as the other implementation.

pybikes/bysykkel.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, tag, meta, feed_url, feed_details_url, key):
+        super(BySykkel, self).__init__(tag, meta)
+        self.feed_url = feed_url
+        self.feed_details_url = feed_details_url
+        self.key = key
+
+    def update(self, scraper=None):
+        if scraper is None:
+            scraper = utils.PyBikesScraper()
+            scraper.headers['Client-Identifier'] = self.key
```
eskerda: This line should go outside of the `None` condition, since we always want it, no matter what scraper object we receive. Sorry if this part is a bit unclear, but makes sense once systems use pybikes

```
scraper = scraper or utils.PyBikesScraper()
scraper.headers['Client-Identifier'] = self.key
```

pybikes/bysykkel.py
```diff
@@ -0,0 +1,71 @@
...
+
+        super(BySykkelStation, self).__init__()
+
+        self.name = info['title']
+
+        self.longitude = float(info['center']['longitude'])
+        self.latitude  = float(info['center']['latitude'])
+
+        self.bikes = info['availability']['locks'] + info['availability']['bikes']
+        self.free = info['availability']['bikes']
```
eskerda: Is it possible these are the other way around?

`bikes` stands for, number of available bikes to get
`free` stands for number of empty spaces to park a bike

pybikes/bysykkel.py
```diff
@@ -0,0 +1,71 @@
...
+        self.extra = {
+            'slots':  info['availability']['locks'],
```
eskerda: Usually we only add slots when we can get the info directly from the canonical source. It means, total number of spaces stations always have. Looking at the info from the API, that would be `number_of_locks`

pybikes/bysykkel.py
```diff
@@ -0,0 +1,71 @@
...
+            'uid': info['id'],
+            'placement': info['subtitle'],
+            'ready': info['ready']
```
eskerda: `ready` key does not seem to appear on the feed (or at least, on the Oslo one). It's probably `in_service` now?

olavxxx:
Update scope of header and changed API for locks/free etc.
Removed the extra metadata that was not uniform across cities.

eskerda:
Looks good, thanks!

Issue #270, opened 2017-07-24
GB vs UK Country code


nicholasshirley:
I'm not sure if this is the right place to raise this, but I was flipping through the data and I noticed that there is one (and only one) bike share company that uses the country code "UK" instead of "GB" as the rest of the United Kingdom companies do. Here is the entry:

```
  {
      "company": [
        "Groundwork",
        "Slough Borough Council",
        "ITS"
      ],
      "href": "/v2/networks/cycle-hire-slough",
      "id": "cycle-hire-slough",
      "location": {
        "city": "Slough",
        "country": "UK",
        "latitude": 51.51135,
        "longitude": -0.591562
      },
      "name": "Cycle Hire"
    },
```

nicholasshirley:
I see something similar is already brought up on #89 though this station is not included in the list.

nicholasshirley:
Fixed with #271

PR #271, opened 2017-07-26
Update country code to match other GB bike shares

nicholasshirley:
Slough Borough Council is the only bike share to
use UK as the country code (Ref issue #270). The other United Kingdom
bike shares appear to use GB. Change only changes country code.


eskerda:
:+1:

PR #272, opened 2017-07-29
Switch Bike Bike Chattanooga to use GBFS

afischer:
Bike Chattanooga supports GBFS, a [standard developed by NABSA](https://github.com/NABSA/gbfs/blob/master/systems.csv), and should be used as the source for bike share information for the system.

The company has also been changed to `Motivate International, Inc`, which replaced Alta and [now operates the system](https://www.motivateco.com/where-we-do-it/chattanooga-tn/).

### DIFF
--- a/pybikes/data/bixi.json
+++ b/pybikes/data/bixi.json
@@ -29,22 +29,6 @@
             "format": "xml"
         },
         {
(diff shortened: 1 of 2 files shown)


eskerda:
Looking good. Thanks!

Issue #258, opened 2017-06-11
Add Whitebikes bike sharing system


andrewshadura:
Please add [Bratislava bike sharing system](http://whitebikes.info/), and possibly more based on the same code: http://opensourcebikeshare.com/

eskerda:
Last I heard from opensourcebikeshare they told me they were planning on supporting the GBFS specification. I guess this is not out there yet.

For reference, this is the current feed http://whitebikes.info/command.php?action=map:markers

andrewshadura:
So, is it difficult to support?

eskerda:
Not at all, seems like a pretty standard JSON feed

edumucelli:
Added it.

PR #281, opened 2017-09-09
Add WhiteBikes Bratislava

edumucelli:
Issue #258

### DIFF
--- a/pybikes/data/open_source_bike_share.json
+++ b/pybikes/data/open_source_bike_share.json
@@ -1,0 +1,17 @@
+{
+    "instances":[
+       {
+            "tag":"whitebikes",
+            "meta":{
+                "latitude":48.149424,
+                "city":"Bratislava",
+                "name":"WhiteBikes",
+                "longitude":17.113922,
+                "country":"SK"
+            },
+            "feed_url":"http://whitebikes.info/command.php?action=map:markers"
+        }
+    ],
+    "system":"opensourcebikeshare",
+    "class":"OpenSourceBikeShare"
+}
--- a/pybikes/open_source_bike_share.py
+++ b/pybikes/open_source_bike_share.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Eduardo Mucelli Rezende Oliveira <edumucelli@gmail.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import json
+
+from pybikes.base import BikeShareSystem, BikeShareStation
+from pybikes import utils
+
+
+class OpenSourceBikeShare(BikeShareSystem):
+
+    meta = {
+        'system': 'OpenSourceBikeShare',
+        "company": ["Open Source Bike Share"]
+    }
+
+    def __init__(self, tag, meta, feed_url):
+        super(OpenSourceBikeShare, self).__init__(tag, meta)
+        self.feed_url = feed_url
+
+    def update(self, scraper=None):
+        scraper = scraper or utils.PyBikesScraper()
+
+        stations = []
+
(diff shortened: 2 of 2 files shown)


pybikes/whitebikes.py
```diff
@@ -0,0 +1,47 @@
...
+
+        data = json.loads(scraper.request(self.feed_url))
+
+        for station in data:
+
+            longitude = float(station['lon'])
+            latitude = float(station['lat'])
+
+            name = station['standName']
+            free = 0
```
eskerda: I recall we have some networks on we set `free` to None, and end up on the API as null. Long time ago we signaled no info on `free` with a -1, but in the end, I think `None` is the best.

pybikes/whitebikes.py
```diff
@@ -0,0 +1,47 @@
...
+            bikes = int(station['bikecount'])
+
+            extra = {
+                'photo': station['standPhoto'],
+                'description': station['standDescription'],
+                'number': int(station['standId'])
```
eskerda: Does `standId` work as `uid` instead of `number`?

pybikes/whitebikes.py
```diff
@@ -0,0 +1,47 @@
...
+            }
+
+            station = BikeShareStation(name, latitude, longitude, bikes, free,
+                                       extra)
+            stations.append(station)
+
+        self.stations = list(stations)
```
eskerda: There should be no need to explicitly create a copy of `stations`, `self.stations = stations` should work.

pybikes/whitebikes.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (C) Eduardo Mucelli Rezende Oliveira <edumucelli@gmail.com>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import json
+
+from pybikes.base import BikeShareSystem, BikeShareStation
+from pybikes import utils
+
+
+class WhiteBikes(BikeShareSystem):
```
eskerda: Can we rename the system to `OpenSourceBikeShare`, since there are other systems that have this same schema. More info at http://opensourcebikeshare.com/

pybikes/data/whitebikes.json
```diff
@@ -0,0 +1,20 @@
...
+            "tag":"whitebikes",
+            "meta":{
+                "latitude":48.149424,
+                "city":"Bratislava",
+                "name":"WhiteBikes",
+                "longitude":17.113922,
+                "country":"SK",
+                "company":[
+                    "Open Source Bike Share"
+                ]
```
eskerda: Once this `company` info is written on the python implementation, there will be no need to put it here.

pybikes/whitebikes.py
```diff
@@ -0,0 +1,47 @@
...
+
+    meta = {
+        'system': 'WhiteBikes'
+    }
+
+    def __init__(self, tag, meta, feed_url, bounding_box=None):
```
eskerda: This system, so far, does not use bounding boxes so there's no need to include it as an init parameter.

edumucelli: True, bad copy n paste!

pybikes/whitebikes.py
edumucelli: Yep, I was not sure whether it was 0, -1, or None.

pybikes/whitebikes.py
edumucelli: Yep, that is correct!

pybikes/whitebikes.py
edumucelli: Yep

edumucelli:
Applied remarks @eskerda

eskerda:
@edumucelli sure, squash them then ready to merge. Looking good!

PR #282, opened 2017-09-11
[Bixi] update Aspen and Melbourne URLs

edumucelli:


eskerda:
Fixed by https://github.com/eskerda/pybikes/pull/400

PR #283, opened 2017-09-11
Fixes to bcycle and smoove

MartinoMensio:
The xpath to the station name for bcycle has changed.
Smoove was having problem because some results had the coordinates as the empty string, causing the whole result to fail. The solution is to skip disabled stations


eskerda:
LGTM, thanks!

PR #292, opened 2017-11-02
Adds open source bike share

eskerda:


eskerda:
@edumucelli Did not want to ask for further changes, you've done enough and good already. This includes your changes + the fix

Issue #299, opened 2018-01-07
New velib


Bahjesaispas:
New update for the .json file in paris. 

The new adress seems to be this one : https://www.data.gouv.fr/fr/datasets/velib-2-disponibilite-en-temps-reel/#_

The old one doesnt work anymore and JcDecaux is remplaced by Smoove.

eskerda:
There's also this feed https://velib-metropole-opendata.smoove.pro/opendata/Velib_Metropole/station_status.json that should be GBFS but doesn't exactly comply with the standard fields.

Phyks:
Hi,

Is anyone already working on this? I can have a look and make a MR to this repo else, as I already have to write some code for the new Velib API.

I have been dumping regularly data from the Velib API for the past 2 years with custom code, and just discovered pybikes. As I have to rewrite my own code anyways, I am thinking about moving to pybikes and contributing this :)

Phyks:
Actually, both API endpoints mentionned above (https://velib-metropole-opendata.smoove.pro/opendata/Velib_Metropole/station_status.json and https://opendata.paris.fr/explore/dataset/velib-disponibilite-en-temps-reel/information/) report only part of the infos (typically, there is no info about which proportion of bikes is manual and which proportion is electric).

https://www.velib-metropole.fr/webapi/map/details?gpsTopLatitude=49&gpsTopLongitude=2.85&gpsBotLatitude=48.7&gpsBotLongitude=2&zoomLevel=19 is the API used by the official website and does contain all the available info, although there is no clear license for these data :/

Bahjesaispas:
I've asked on twitter if we can use the official website information instead the open data. Waiting for an answer...

Phyks:
Same here for a month now :/

eskerda:
It's alright to use it. I would prefer to use the GBFS feed to avoid creating yet another implementation, but if it's missing data then why not go full in and get the feed with more info

eskerda:
> Anyways, that's the best data available for now, so I'd rather use them as well. How would you handle the electric bikes vs standard bikes thing? I guess bikes should be the sum of the two, but how would you keep track of this detailed info (having citybik.es and apps that use it in mind)?

Other systems already have this. Not really happy on the syntax, but I guess it works. For the `smartbike` system, we have:
* `has_ebikes` => `true`
* `extra['ebikes']` => number of e-bikes on the station

Phyks:
#300 should be implementing this. Let me know if you have any comments about this PR. :)

PR #300, opened 2018-01-14
Add a Smovengo class to handle the new Velib' system in Paris

Phyks:
As discussed in https://github.com/eskerda/pybikes/issues/299. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ these cases, the instance factory can take an extra API key parameter.
 
 ```python
 >>> key = "This is not an API key"
->>> velib = pybikes.get('velib', key)
+>>> dublinbikes = pybikes.get('dublinbikes', key)
 ```
 
 Note that pybikes works as an instance factory and, choicely, instances can be
--- a/pybikes/base.py
+++ b/pybikes/base.py
@@ -76,7 +76,7 @@ class BikeShareSystem(object):
         - General (companies):
             - JCDecaux, ClearChannel
         At the same time, these classes can extend their base class,
-        for example, Velib extends JCDecaux extends BikeShareSystem.
+        for example, Velib extends Smovengo extends BikeShareSystem.
 
         This class might or not have METADATA assigned, usually intended
         for specific cases. This METADATA is dict / json formatted.
--- a/pybikes/data/cyclocity.json
+++ b/pybikes/data/cyclocity.json
@@ -1,77 +1,66 @@
 {
-    "system": "cyclocity", 
+    "system": "cyclocity",
     "class": {
         "CyclocityWeb": {
             "instances": []
-        }, 
+        },
         "Cyclocity": {
             "instances": [
                 {
-                    "tag": "cyclic", 
+                    "tag": "cyclic",
                     "meta": {
-                        "latitude": 49.44323199999999, 
-                        "country": "FR", 
-                        "name": "cy'clic", 
-                        "longitude": 1.099971, 
+                        "latitude": 49.44323199999999,
+                        "country": "FR",
+                        "name": "cy'clic",
+                        "longitude": 1.099971,
                         "city": "Rouen"
-                    }, 
+                    },
                     "contract": "Rouen"
-                }, 
+                },
                 {
-                    "tag": "velib", 
+                    "tag": "velo",
                     "meta": {
-                        "latitude": 48.856614, 
-                        "country": "FR", 
-                        "name": "Velib", 
-                        "longitude": 2.3522219, 
-                        "city": "Paris"
-                    }, 
-                    "contract": "Paris"
-                }, 
-                {
-                    "tag": "velo", 
-                    "meta": {
-                        "latitude": 43.604652, 
-                        "country": "FR", 
-                        "name": "V\u00e9l\u00f4", 
-                        "longitude": 1.444209, 
+                        "latitude": 43.604652,
+                        "country": "FR",
+                        "name": "V\u00e9l\u00f4",
+                        "longitude": 1.444209,
                         "city": "Toulouse"
-                    }, 
+                    },
                     "contract": "Toulouse"
-                }, 
+                },
                 {
-                    "tag": "veloh", 
+                    "tag": "veloh",
                     "meta": {
-                        "latitude": 49.6301025, 
-                        "country": "LU", 
-                        "name": "Veloh", 
-                        "longitude": 6.1520185, 
+                        "latitude": 49.6301025,
+                        "country": "LU",
+                        "name": "Veloh",
+                        "longitude": 6.1520185,
                         "city": "Luxembourg"
-                    }, 
+                    },
                     "contract": "Luxembourg"
-                }, 
+                },
                 {
-                    "tag": "valenbisi", 
+                    "tag": "valenbisi",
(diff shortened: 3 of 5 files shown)


eskerda:
@Phyks thanks for the contribution!

Do you have any insights on the Paris bike share? This feed is reporting 172 stations (a really small number compared to before), and many of them have status `Work in progress` or/and report 0 bikes and 0 slots.

If this feed is proper, I would suggest filtering stations that are in progress, and leaving the status field at all until we know if there's any other state other than `Work in progress` and `Operative`.

Other than that, LGTM

Phyks:
Thanks for the review!

> Do you have any insights on the Paris bike share? This feed is reporting 172 stations (a really small number compared to before), and many of them have status Work in progress or/and report 0 bikes and 0 slots.

Sure (I live in Paris :). Basically, it used to be operated by Cyclocity but the contract ended on 1st of January and Velib is now being operated by Smovengo.

There are about 3k stations around Paris, but they are very late in the process of conversion. So far, there is indeed only a handful of stations operative (and some extra which are currently work in progress, as reported by the API). The number are (sadly) correct :/

The feed I use is basically the feed powering [the official map](https://www.velib-metropole.fr/map#/) (then, license is not very clear).

> I would suggest filtering stations that are in progress, and leaving the status field at all until we know if there's any other state other than Work in progress and Operative.

I can filter out the "Work in progress" stations, that seems reasonable. However, I would rather keep the status field to have an easy way to track whether there are other status. Note that keeping "Work in progress" station may be of interest to be able to know whether a station is already operational / in work and the expected date (in the `dueDate` field). This might be of interest in the future if some work is planned for a given station.


P.S.: Are you generating datasets of data fetched by the citybik.es API? I've been [dumping the status of all Velib stations](https://pub.phyks.me/datasets/velib/) for around two years now, every 5 mins and was planning to move my "fetch from API" script to this PyBikes module. But if citybik.es is also providing dump, that would be duplicate :)

eskerda:
> I can filter out the "Work in progress" stations, that seems reasonable. However, I would rather keep the status field to have an easy way to track whether there are other status. Note that keeping "Work in progress" station may be of interest to be able to know whether a station is already operational / in work and the expected date (in the dueDate field). This might be of interest in the future if some work is planned for a given station.

Actually you are right. Let's have them, but we should somehow signal that we know this is not a valid station. In some providers, we set `extra.status` to `online`, and in others we have the field `extra.online` as a boolean. It's a mess that we have both, but I think the most consistent thing is declaring too an `extra.online` based on `Operative`. For what it matters, anything not `Operative` is going to be an `online? false`. Not sure if this makes sense to you.

> P.S.: Are you generating datasets of data fetched by the citybik.es API? I've been dumping the status of all Velib stations for around two years now, every 5 mins and was planning to move my "fetch from API" script to this PyBikes module. But if citybik.es is also providing dump, that would be duplicate :)

Tried for a while, but the amount of data we were generating was too much for our budget. Do you mind if I share your velib dumps any time someone sends an email about them?

Bahjesaispas:
I live in Paris too and I can confirm that velib with Jcdecaux doesn't work anymore.
All the station are closed since the 31 December and now we just have the new velib (by Smoove). And the official app doesn't work on linage os without the Gapps...

Phyks:
> As discussed let's add extra.online field as a boolean wether status is Operative or not

That totally makes sense, it should be updated now. I added also a check to ensure the fetched stations are indeed valid (as I saw a "Zzz delete me" station appear this morning, which is obviously a test case).

> Tried for a while, but the amount of data we were generating was too much for our budget.

Erf :/ Was the storage part the issue? Maybe there are some ways to offload the storage to another provider (Google Drive?)? I know arXiv uses a "pay bandwidth to download" model: https://arxiv.org/help/bulk_data_s3. Shouldn't cost them much to keep the data online, but people pay a small fee to download the dataset from Amazon (rather than arXiv paying for hosting).


> Do you mind if I share your velib dumps any time someone sends an email about them?

No problem, that's why they are online in the first place! :)


Last remaining concern to mention is that in the online API, there are some duplicates (actually one at the moment). Look at `10006` in https://www.velib-metropole.fr/webapi/map/details?gpsTopLatitude=49&gpsTopLongitude=2.85&gpsBotLatitude=48.7&gpsBotLongitude=2&zoomLevel=19, you will see that there is one operative station and one WIP station, but they should be the same. Not sure whether it is PyBikes job to make a choice here or not? (I don't know what is the reality :/)

eskerda:
> Last remaining concern to mention is that in the online API, there are some duplicates (actually one at the moment). Look at 10006 in https://www.velib-metropole.fr/webapi/map/details?gpsTopLatitude=49&gpsTopLongitude=2.85&gpsBotLatitude=48.7&gpsBotLongitude=2&zoomLevel=19, you will see that there is one operative station and one WIP station, but they should be the same. Not sure whether it is PyBikes job to make a choice here or not? (I don't know what is the reality :/)

At the moment, at least on my side, there's no problem. The idea behind storing the unique ids is to generate better and more reliable ids, but right now another method is being used.

eskerda:
@Phyks thanks again for contributing! Let's see how the feed goes with time.

Bahjesaispas:
Tryed. The electric bike are not supported or they just look the same on the app.

eskerda:
@Bahjesaispas you are on the wrong repo

Phyks:
Thanks for merging this!

> Last remaining concern to mention is that in the online API, there are some duplicates

Small update on this, apparently (from people checking on the ground) the station "10006" is already working and the WIP duplicate is not a real station.


> Do you intend to add it for all cities? That'd be great. This discussion made me figure out that the extra.status field is not consistent. So there is no way today to know if a station is opened or not, is there?

Might be worth moving it out of the `extra` field, then? Standardizing `ebikes` and `status` might be a good idea. But this is not at all the goal of this PR :)

> Is the smoove.pro url official? How did you find it? 

Totally not, but it is the only "complete" and usable set of data available out there. There is no explicit license for these data, and it is coming from the official website (http://velib-metropole.fr/) (it is the API used behind the map).

I know that http://velib.nocle.fr/ uses it as well and has been retweeted by the official Velib Twitter account, so I guess we can use this data for now.

The official opendata are https://opendata.paris.fr/explore/dataset/velib-emplacement-des-stations/ and https://opendata.paris.fr/explore/dataset/velib-disponibilite-en-temps-reel/ but it's basically a mess, not usable (inconsistencies and a lot of missing data).

sthocs:
> Might be worth moving it out of the extra field, then? Standardizing ebikes and status might be a good idea. But this is not at all the goal of this PR :)

Sure, I was using the occasion to jump into the conversation and suggest the thing to @eskerda ;)

> it is coming from the official website (http://velib-metropole.fr/) (it is the API used behind the map)

I wanted to know about the one coming from the smoove.pro website, in pseudo GBFS (which has been announced as the chosen format). But yeah I'll use the website API too for now.

Phyks:
> I wanted to know about the one coming from the smoove.pro website, in pseudo GBFS (which has been announced as the chosen format). But yeah I'll use the website API too for now.

Oh, sure, I did not understand this. You should ask @eskerda, he/she is the one who found it :)

PR #291, opened 2017-11-02
add le velo star Rennes, FR

BenSto:
feel free to have a look. potentially closes #235

PR #305, opened 2018-01-31
Fix keolis STAR parser

eskerda:
@BenSto Problem with https://github.com/eskerda/pybikes/pull/291 was that it was only getting 10 records, plus some style issues. This PR fixes them


eskerda:
Alright @BenSto , all merged! Thanks for the initial contribution. Sorry it took long. I saw that it was not getting the full results and completely forgot about it.

Issue #311, opened 2018-03-21
Use GBFS feed for Bixi Montréal system


f8full:
The feed is available here : https://api-core.bixi.com/gbfs/gbfs.json

source : https://www.bixi.com/en/open-data

f8full:
WIP : https://github.com/f8full/pybikes/commits/master

f8full:
And the PR (sry if there was a way to automagically link everything)
https://github.com/eskerda/pybikes/pull/312

eskerda:
Great! Issues that come together with PR that fix them are the best, so this is greatly appreciated. Thanks!

PR #312, opened 2018-03-21
Changes for #311

f8full:
My first attempt at contributing. (Full disclosure https://github.com/f8full/findmybikes)

I tried to resolve the issue I opened regarding the use of a GBFS feed instead of an XML one for the Montréal Bixi system in Québec, Canada

@eskerda obviously I didn't know I could add a comment before submitting a pull request (it also IS my first github PR #proud), sry for the double tagging !

Of course I'll undertake any further required changes or tests, please let me know

Cheers,

F.


pybikes/bixi.py
```diff
@@ -1,211 +0,0 @@
...
-            'lastCommWithServer': data['lastCommWithServer'],
-            'installed': utils.str2bool(data['installed']),
-            'locked': utils.str2bool(data['locked']),
-            'installDate': data['installDate'],
-            'removalDate': data['removalDate'],
-            'temporary': utils.str2bool(data['temporary']),
-            'public': utils.str2bool(data['public']),
-            'latestUpdateTime': data['latestUpdateTime']
-        }
-        return station
```
eskerda: This file cannot be deleted yet. Other systems (these are present on [`pybikes/data/bixi.json`](https://github.com/f8full/pybikes/blob/7f043af9940802397d62ad584dedad772df0964b/pybikes/data/bixi.json)) are using it.

pybikes/data/bixi.json
```diff
@@ -1,17 +1,5 @@
...
-            "meta": {
-                "latitude": 45.5086699,
-                "city": "Montreal, QC",
-                "name": "Bixi",
-                "longitude": -73.55399249999999,
-                "country": "CA"
-            },
-            "feed_url": "https://montreal.bixi.com/data/bikeStations.xml",
-            "format": "xml"
-        },
```
eskerda: :+1:

pybikes/data/gbfs.json
```diff
@@ -29,6 +29,21 @@
...
+                "city": "Montréal, QC",
+                "name": "Bixi Montréal",
+                "country": "CA",
+                "company": [
+                    "Motivate International, Inc.",
+                    "PBSC Urban Solutions"
+                ]
+            },
+            "feed_url": "https://api-core.bixi.com/gbfs/gbfs.json"
+        },
```
eskerda: :+1:

f8full:
I think that should do the trick, thks GitHub desktop !

pybikes/bixi.py
f8full: Ooopsy
I learnt my lesson
https://github.com/f8full/pybikes/commit/7c3cdf9aecf22975805a7d2570e4f70edbb8dfc1

someday I may write a parser for this file that would feed pybikes
https://github.com/NABSA/gbfs/blob/master/systems.csv

cheers,

eskerda:
Looks good. Now the problem with that revert commit is that the whole git blame on `bixi.py` points to you, for no particular reason.

Revert commits are good for reverting things that have already landed master or some downstream projects, but in this case, being a PR, there's no reason for not rewriting history and doing a force push.

There are many ways you can do this if you want. **If not, I can cherry-pick the needed commits on a branch and merge it myself**, but I think it's always interesting to learn git the hard way :) So, here's how you can get that fixed:

## (Easy) Reset hard to origin/master and cherry pick commits
Since you are doing these changes on your master, you cannot reset hard into origin master, but you can rewind the head to the latest commit here
```
# This will put your head the same as the current pybikes head
git reset --hard 9d35934a3155c5b709c4039481ca31463006101f 
# The changes you did, however, are still somewhere so you can cherry pick them
git cherry-pick e6741000d5bbb8b38e44294e53975c31b28086e9
git cherry-pick 1d12a62377e018c3aa1caf2dad59e12867b8b84c
```
Now force push your repo, this PR should get updated. 
```
git push -f origin master
```

## (Medium) Use git rebase interactive
Simply
```
git rebase -i HEAD~4
```
Now it will show a screen with 4 commits, you want to remove from there the ones you are not interested into (the last two lines). This is the reason why you usually want to give git commits descriptive messages.
Again, force push that, and this PR should get updated.

# A note on working on master
Usually, avoid working on master, and instead work on a feature branch. A branch can be created like
```
git checkout -b my-feature-branch-name
```
After this, you are on a different branch, you can play on that branch, commit things, and whenever you go back to master, things will be like they were before
```
git checkout master
```
You can list the branches that you have by
```
git branch
```
You can switch between branches the same way
```
git checkout my-feature-branch-name
```
You would then push that feature branch onto your repo, and then follow the usual way to create a pull request.

The idea behind this is, usually, that you want master to be the same (or not up to date) to mine, and try to avoid diverging from there. Also, unless justified, you should never force push on master, since you could leave other people that had your master on a detached state.

f8full:
Just opened #313 by cranching correctly as you described. I'll fix my repo using the great guidance you sahred, thks for that !

PR #315, opened 2018-04-08
Update some systems from Brazil to use GBFS feed

igormatos:


eskerda:
Looks good, thanks @igormatos!

Issue #318, opened 2018-05-13
Add Mibici Guadalajara


homtec:
feed-url: https://guad.publicbikesystem.net/ube/gbfs/v1/

Guadalajara, MX
network: mibici_guadalajara
web: https://www.mibici.net/en/

Altonss:
This issue can be closed as it has been added successfully :)

Altonss:
To be close @eskerda ;)

PR #319, opened 2018-05-13
Added Mibici Guadalajara 

homtec:
#318


eskerda:
@homtec looks good, thanks! :bike:

PR #321, opened 2018-05-22
Add initial support for Trondheim

odinuge:
The new city bike system in Trondheim use Urban Infrastructure
Partners new graphql API. The old api (used in Oslo etc.) does not
work for Trondheim, and the new graphql one is only for Trondheim.

GraphQL for the win! :smile: 

https://www.trondheimbysykkel.no/


pybikes/bysykkelgraphql.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+import json
+
+import requests
```
eskerda: Unused import

pybikes/bysykkelgraphql.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.bikes = len(
+            filter(
+                lambda dock: dock['availabilityState'] == 'vehicle_available' or dock['availabilityState'] == 'available',
+                info['docks']))
+        self.free = len(
+            filter(
+                lambda dock: dock['availabilityState'] == 'vehicle_available',
+                info['docks']))
+
```
eskerda: I do not understand the logic behind `vehicle_available` both accounting for `self.bikes` and `self.free`.

* `self.bikes`: Number of available bikes
* `self.free`: Number of available parking spaces
(Additionally when available)
* `self.extra['slots']`: Total number of spaces on stations (no matter the status or if it is used)

pybikes/bysykkelgraphql.py
```diff
@@ -0,0 +1,92 @@
...
+        self.extra = {
+            'id': info['id'],
```
eskerda: should be `uid`

pybikes/data/bysykkelgraphql.json
```diff
@@ -0,0 +1,19 @@
+{
+    "instances":[
+        {
+            "tag":"trondheim-bysykkel",
+            "meta":{
+                "latitude":63.4307238,
+                "country":"NO",
+                "name":"Trondheim Bysykkel",
```
eskerda: Since we already have the city name, only use `Bysykkel` as the name

pybikes/bysykkelgraphql.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+import json
+
+import requests
+
+from . import utils
+from .base import BikeShareStation, BikeShareSystem
+
+
+class BySykkelGraphql(BikeShareSystem):
```
eskerda: ~~Is it `Bysykkel` or `BySykkel`?~~ EDIT: My bad, I see that our current code for bysykkel calls it BySykkel. Still, if you do not mind, I would set it as `BysykkelGraphql` and I will change the other one too

Also, the file name should be `bysykkel_graphql.py`

pybikes/bysykkelgraphql.py
```diff
@@ -0,0 +1,92 @@
...
+
+    authed = False
+
+    meta = {
+        'system': 'BySykkelGraphql',
```
eskerda: This name is only explanatory for the codename of the service being used, so should be something informative that helps correlate systems through different cities. This could be `Bysykkel GraphQL`. It's not really used anywhere, actually.

pybikes/data/bysykkelgraphql.json
```diff
@@ -0,0 +1,19 @@
...
+                "longitude":10.3936781,
+                "city":"Trondheim",
+                "city_code":"TRD",
+                "systemId": "trondheim"
+            },
+            "feed_url":"https://core.urbansharing.com/public/api/v1/graphql"
+        }
+    ],
+    "system":"bysykkelgraphql",
```
eskerda: If you change the filename, this will have to be `bysykkel_graphql`, and the filename will be `bysykkel_graphql.json`

pybikes/data/bysykkelgraphql.json
```diff
@@ -0,0 +1,19 @@
...
+    "class":"BySykkelGraphql"
```
eskerda: This will change to `BysykkelGraphql`

pybikes/bysykkelgraphql.py
```diff
@@ -0,0 +1,92 @@
...
+            'query': query,
+            'operationName': 'dockGroups',
+            'variables': variables
+        }
+
+        data = json.loads(
+            scraper.request(
+                self.feed_url,
+                data=json.dumps(gql_data).encode('utf-8'),
+                method='POST'))
```
eskerda: Does this request randomly fail to you too? I have had 2/5 of success running it. Others return a 5xx from their service

eskerda:
@odinuge thanks for the PR. Definitely, graphql for the win!

Please review the requested changes, and feel free to discuss them.

pybikes/data/bysykkelgraphql.json
```diff
@@ -0,0 +1,19 @@
...
+            "meta":{
+                "latitude":63.4307238,
+                "country":"NO",
+                "name":"Trondheim Bysykkel",
+                "longitude":10.3936781,
+                "city":"Trondheim",
+                "city_code":"TRD",
+                "systemId": "trondheim"
+            },
+            "feed_url":"https://core.urbansharing.com/public/api/v1/graphql"
```
eskerda: Seems to me this URL is going to be common to any system on this feed, and the only thing changing would be the `systemId` right?

I would move `systemId` to the JSON level where `feed_url` currently is, and move `feed_url` directly into the python code.

odinuge: Yes, I think so. Will change it! :smile:

pybikes/bysykkelgraphql.py
odinuge: Hmm. Works for me every time. Since you get a `5XX` response, I guess they get a message about the error; hopefully resulting in them fixing the issue.

odinuge:
This API looks cool. They even provide live tracking of each bike (may be just for superusers, because it doesn't work even with auth). Each bike contains a small computer with GPS and other sensors. More info in this article; in [norwegian here](https://nrkbeta.no/2018/05/27/vi-kunne-ogsa-oversvomt-oslos-gater-med-20-000-sykler/
) and using google translate in [english here](https://translate.google.no/translate?hl=no?sl=no&tl=en&u=https%3A//nrkbeta.no/2018/05/27/vi-kunne-ogsa-oversvomt-oslos-gater-med-20-000-sykler/).


Small intro about how (i think) the API works:

Each `dockGroup` is one physical location, with multiple bikes. Each `dockGroup` has multiple `dock`s, and each dock can hold one bike.

`dockGroup`:
- `state`: `active` if the dockGroup is action, else `i don't know`

`dock`:
- `state`: (deprecated - should remove the usage)
    - `broken`: The `dock` is broken, or that it contains a broken bike
   - `available`: Available
- `availabilityState`:
    - `available`: Free spot
    - `vehicle_available`: Bike available
    - `unavailable`: Out of order (broken, broken bike etc.)
- `vehicleDocked`: `boolean`, is bike docked.


The official map, that is using the same api, can be found here: https://www.trondheimbysykkel.no/stasjoner

pybikes/data/bysykkel_graphql.json
```diff
@@ -0,0 +1,18 @@
...
+        {
+            "tag":"trondheim-bysykkel",
+            "meta":{
+                "latitude":63.4307238,
+                "country":"NO",
+                "name":"Bysykkel",
+                "longitude":10.3936781,
+                "city":"Trondheim",
+                "city_code":"TRD",
+                "systemId": "trondheim"
```
eskerda: If you move `systemId` outside of `meta`, it will be available on initialization, see: [pybikes/nextbike.py](https://github.com/eskerda/pybikes/blob/master/pybikes/nextbike.py#L31) and [pybikes/data/nextbike.json](https://github.com/eskerda/pybikes/blob/master/pybikes/data/nextbike.json#L4-L12).

pybikes/bysykkel_graphql.py
```diff
@@ -0,0 +1,90 @@
...
+    authed = False
+
+    meta = {
+        'system': 'Bysykkel GraphQL',
+        'company': ['Urban Infrastructure Partner'],
+    }
+
+    def __init__(self, tag, meta):
+        super(BysykkelGraphql, self).__init__(tag, meta)
+        self.feed_url = BYSYKKEL_GRAPHQL_URL
```
eskerda: Once you get the `systemId` here you can leave BYSYKKEL_GRAPHQL_URL outside the class (see comment about the json data file).

eskerda:
> This API looks cool. They even provide live tracking of each bike (may be just for superusers, because it doesn't work even with auth). Each bike contains a small computer with GPS and other sensors. [...]

Cool! We do not support single bikes for now, but it's clearly on the roadmap somewhere.

pybikes/bysykkel_graphql.py
odinuge: Aha, should be good to go now.

eskerda:
Looking good @odinuge! Many thanks for your contribution 🚲

Issue #190, opened 2016-09-04
Support for Bikes


bcaller:
Something to decide when making the GBFS parser in #140 :
How do we deal with [free bikes that are not at a station](https://github.com/NABSA/gbfs/blob/master/gbfs.md#free_bike_statusjson)? Should we add them to the `stations` list? (probably not)

As an example, http://velogo.ca/opendata/free_bike_status.json lists bikes all over the city which can be unlocked and rented.

``` JSON
{
  "bike_id": "bike_2952",
  "name": "2239",
  "lon": -75.72233333333334,
  "lat": 45.435476666666666,
  "is_reserved": 0,
  "is_disabled": 0
}
```

eskerda:
This is also relevant for nextbike and the call a bike system

jbg:
In my city there is one provider with stations, and three providers that only have free bikes. As far as I can see only the provider with stations is mapped. Is there support in pybikes for free bikes yet? If not, is there any existing discussion going on about how to support them?

eskerda:
> In my city there is one provider with stations, and three providers that only have free bikes. As far as I can see only the provider with stations is mapped. Is there support in pybikes for free bikes yet? If not, is there any existing discussion going on about how to support them?

This is the only ongoing discussion AFAIK. There's no support yet. Support will depend on the different operators. Some might decide to release the data openly, whilst other do not want to be scraped, with strong rate limiting, account-only token, etc. See: https://github.com/ubahnverleih/WoBike

hackerb9:
Please do handle dockless systems. I hate running a proprietary app on my phone just to rent a bike.

Or am I misunderstanding what is possible? Are the APIs only able to locate bikes, but not unlock them? If that's the case, it still makes sense to add dockless systems because, as @jbg pointed out, some free floating bikes do not need to be unlocked to be used.

syssi:
+1 for free floating nextbikes.

eskerda:
https://github.com/eskerda/pybikes/pull/792

PR #324, opened 2018-06-02
Removed standalone bike filtering

jakobchwastek:
Why shouldn't we deliver the infos about the standalone bikes in the Nextbike networks. If a user doesn't need the information one can still filter it by their own.


eskerda:
Because an station is not a bike, so if we do not offer a clear distinction between them it's just confusing. Stations do not change position. Please refer to https://github.com/eskerda/pybikes/issues/190

PR #327, opened 2018-06-18
[nextbike] Updated Polish networks

adamdanielczyk:
This PR updates Polish nextbike stations according to public json api: https://api.nextbike.net/maps/nextbike-live.json?list_cities=1


eskerda:
LGTM, thanks @adamdanielczyk !

Also worth checking the json api, thanks for that info too

PR #331, opened 2018-08-05
[nextbike] Networks update

adamdanielczyk:
This commit updates some of the Nextbike networks.

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -1,7 +1,7 @@
 {
     "instances": [
         {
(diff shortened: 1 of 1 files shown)


eskerda:
I am going to accept this PR as is. There are way too many networks to check. On a side note, nextbike systems keep appearing and disappearing from their API, so it's a real pain in the ass to keep them up to date.

eskerda:
Thanks for the contribution and house keeping nextbike @adamdanielczyk!

Issue #91, opened 2015-11-30
bicipalma not responding


lemieuxs:
When I try to run unittest_pybikes.py When it gets to bicipalma network the update doesn't work.
However when I query http://api.citybik.es/v2/networks/bicipalma I get a response...

Are you aware of this issue, is it temporary ?

eskerda:
You can see a timestamp of when bicipalma was last updated `2015-10-12T11:21:17.712Z` on the feed.

Unfortunately, it seems even the official map is not working, neither has been for a long time:

This is the feed http://83.36.51.60:8080/eTraffic3 linked at http://bicipalma.palmademallorca.es/portal/BICIPALMA/home.jsp?codResi=9 (MAPA DE LES ESTACIONS)

eskerda:
Nothing we can do here, except maybe investigate alternative ways of getting this info

lemieuxs:
Thx for the fast answer !

PR #333, opened 2018-09-11
[bicipalma] Fix system

eskerda:
If at some point we get deauth errors it's possible to refresh the auth token.


eskerda:
Fixes https://github.com/eskerda/pybikes/issues/91

PR #342, opened 2018-10-22
[decobike] Fix San Diego Feed Url

### LINKED ISSUE #306: San Diego DecoBike showing station info for Miami's City Bike
Hi...I came across an issue that I thought was due to my coding, but it turns out every single Lat/Lon data for San Diego's Decobike is wrong.  The vendor info for Decobike (city, lat/lon) all appears correct - its just the individual station Name/Addresses and Lat/Lon all appear to be valid bike hub addresses in Miami for City Bike locations.

Every lat/lon is pretty much the same:   The network call is returning 99 stations, I sort them in order of a calculated distance to where a user's starting point is.  Below are the first and last of those 99 sorted by lat/lon distances to the user:

```
station: 0,
latitude:25.76624
longitude:-80.20121
name:"569 - NW 1st St & 1st Ave"
...
...
station:98,
latitude:25.7653275
longitude:-80.1313248
name:"101 - Inside South Pointe Park"
```

How can we get this data corrected?  Is this something being fed by Decobike and they need to be contacted or is this possibly merged data within citybik.es?

rtran9:
Fixes #306

### DIFF
--- a/pybikes/data/decobike.json
+++ b/pybikes/data/decobike.json
@@ -19,7 +19,7 @@
                 "longitude":-117.1610838,
                 "country":"US"
             },
-            "feed_url":"http://www.decobike.com/sandiego/playmoves.xml",
+            "feed_url":"http://www.discoverbikesandiego.com/DiscoverSanDiegoLocations.xml",
             "tag":"decobike-san-diego"
         }
     ],


eskerda:
@rtran9 Thanks for fixing that issue!

PR #344, opened 2018-10-23
[nextbike] Networks update

adamdanielczyk:


eskerda:
@adamdanielczyk A lot of these systems report back 1 station only: Ahmedabad, Bukarest, Birmingham, ... Are they worth adding?

Other than that, focsani reports 0 stations, salzburg, zalec and polzela fail

eskerda:
All networks check fine. Thanks @adamdanielczyk!

I have this script although it's not perfect and might fail

```bash
NEW_STUFF=`git diff -U0 master | grep '^[+-]' | grep -Ev '^(--- a/|\+\+\+ b/)' | grep tag | awk '{print $3}' | sed 's/[\",]//g' | uniq`
echo "Found new systems: $NEW_STUFF"
pip install .
for i in $NEW_STUFF; do
    python tests/unittest_pybikes.py "TestSystems.test_$i"
done
```

PR #348, opened 2018-10-31
Abort tests if anything fails

eskerda:
PR should be failing https://github.com/eskerda/pybikes/pull/348


eskerda:
Okay, at least we got our fails here: https://travis-ci.org/eskerda/pybikes/jobs/448897758

eskerda:
Something something deprecated use github apps yada.

eskerda:
Now it works... so something temporal. Let's merge once this runs

PR #349, opened 2018-11-01
Add Föli bikes in Turku, Finland

ZeiP:
The patch adds support for the Föli-fillari-system by Turku regional traffic in Turku, Finland. It uses the Digitransit API.

### DIFF
--- a/pybikes/data/otp.json
+++ b/pybikes/data/otp.json
@@ -20,6 +20,21 @@
                 "latitude":60.16985569999999
             },
             "feed_url":"https://api.digitransit.fi/routing/v1/routers/hsl/bike_rental"
+        },
+        {
+            "tag":"foli",
+            "meta":{
+                "city":"Turku",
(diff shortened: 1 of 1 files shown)


eskerda:
Looks good, many thanks @ZeiP!

eskerda:
@ZeiP I assume the system is not run by nextbike anymore? In any case, I prefer this feed than using the one by nextbike.

eskerda:
Deleted nextbike entry at https://github.com/eskerda/pybikes/commit/01a1aff44690a1963f987b31a62f3ee91c3faba9

ZeiP:
It actually is run by Nextbike, I have to admit I missed the fact that Nextbike API was already listed in the service. I do still think it's better to use the Digitransit API, because that's the one providing the official route guide and application data; I suppose the data should be the same but this way the integration works even if/when the provider is changed.

PR #352, opened 2018-11-20
Fix Melbourne Bike Share

carterbush:
The feed url and data format for Melbourne Bike Share seem to have changed, resulting in no data being scraped. I've moved it from `bixi.json` to `gbfs.json`, and updated the `feed-url` so it correctly pulls the data.


eskerda:
Hey @carterbush! Thanks for the contribution and sorry for taking long to look at it.

Looks good!

Issue #294, opened 2017-11-08
Add Gira and TCBikes in Lisbon, Portugal


eriickaraujo:
Gira - https://www.gira-bicicletasdelisboa.pt/descobre-as-estacoes/
TCBikes - http://www.tcbarreiro.pt/tcbikes
Mobicascais - https://www.mobicascais.com/rede-mobi/cat/cat_1

martgnz:
TCBikes seems to be temporarily suspended and I can't find a public map for Mobicascais, only a locked-down app.

PR #354, opened 2019-01-14
Added support for the Gira system in Lisbon

miguel71:
Partially fixes https://github.com/eskerda/pybikes/issues/294


eskerda:
Looking good @miguel71, thanks!

PR #355, opened 2019-01-20
Update Bizi Zaragoza API url and format

alejandroscf:
As the URL of the has changed, it no longer works, so I updated it to the new location of the json file and changed the format in order to make it work.


eskerda:
Thanks @alejandroscf ! Looks good :biking_man:

Issue #266, opened 2017-06-22
Bicincitta - Look what happened to these feeds


eskerda:
```
ERROR - {u'meta': {u'latitude': 41.125278, u'city': u'Bari', u'name': u'Bari in Bici', u'longitude': 16.866667, u'country': u'IT'}, u'tag': u'bariinbici', u'system_id': 9}
ERROR - {u'meta': {u'latitude': 45.768611, u'city': u'Bassano del Grappa', u'name': u'Bassano del Grappa', u'longitude': 11.736667, u'country': u'IT'}, u'tag': u'bassanodelgrappa', u'system_id': 45}
ERROR - {u'meta': {u'latitude': 44.694224, u'city': u'Bra', u'name': u'Bra', u'longitude': 7.854795, u'country': u'IT'}, u'tag': u'bra', u'system_id': 14}
ERROR - {u'meta': {u'latitude': 45.73626, u'city': u'Cant\xf9', u'name': u'BICIamo', u'longitude': 9.12569, u'country': u'IT'}, u'tag': u'biciamo', u'system_id': 69}
ERROR - {u'meta': {u'latitude': 44.983499, u'city': u'Casalmaggiore', u'name': u'Casalmaggiore', u'longitude': 10.419245, u'country': u'IT'}, u'tag': u'casalmaggiore', u'system_id': 40}
ERROR - {u'meta': {u'latitude': 44.7620048, u'city': u'Novi Ligure', u'name': u'Novi Ligure', u'longitude': 8.7858956, u'country': u'IT'}, u'tag': u'noviligure', u'system_id': 22}
ERROR - {u'meta': {u'latitude': 45.77976460000001, u'city': u'Oderzo', u'name': u'Oderzo', u'longitude': 12.4973111, u'country': u'IT'}, u'tag': u'oderzo', u'system_id': 49}
ERROR - {u'meta': {u'latitude': 43.9303475, u'city': u'Pistoia', u'name': u'PistoiainBici', u'longitude': 10.9078587, u'country': u'IT'}, u'tag': u'pistoiainbici', u'system_id': 4}
ERROR - {u'meta': {u'latitude': 43.6625156, u'city': u'Pontedera', u'name': u'Pontedera', u'longitude': 10.6363586, u'country': u'IT'}, u'tag': u'pontedera', u'system_id': 64}
[...]

PR #359, opened 2019-01-24
Add support for new bicincitta system

eskerda:
Things have "improved" a lot.

1. Move all systems to new system
2. Keep old system for compatibility with one (?) system

Related: https://github.com/eskerda/pybikes/issues/266


eskerda:
We should try to include the systems on https://github.com/eskerda/pybikes/pull/198 too

PR #361, opened 2019-02-11
Update BikeSantos data

mmmaia:
Update BikeSantos URL and move to SambaNew


eskerda:
Perfect @mmmaia , thanks!

PR #362, opened 2019-02-13
Deal with GBFS feeds with the wrong protocol

mmmaia:
Workaround for GBFS publishers that have feed urls with a protocol different from the main one, for example, https://saopaulo.publicbikesystem.net/ube/gbfs/v1/

### DIFF
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -708,7 +708,8 @@
                     "PBSC Urban Solutions"
                 ]
             },
-            "feed_url": "https://rec.publicbikesystem.net/ube/gbfs/v1/"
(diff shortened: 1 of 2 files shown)


eskerda:
This solution seems very hacky. I mean, it's their fault for not handling the redirect. We either move it to a separate "hacks" file (I can help with that), or add a "force https flag" on gbfs data files (can also help)

mmmaia:
Totally agree. I tried to contact them, but no response yet. I like the "force https flag" idea, will try it.

mmmaia:
@eskerda changed to the implementation to `force_https` flag. WDYT?

eskerda:
> @eskerda changed to the implementation to `force_https` flag. WDYT?

Very straightforward, I like it. thanks!

PR #365, opened 2019-02-18
Fix CicloSampa

mmmaia:
- change data URL
- load data from JSON


pybikes/ciclosampa.py
```diff
@@ -28,31 +27,30 @@ def update(self, scraper=None):
...
+        self.longitude = float(data['longitude'])
+        self.bikes = int(data['bikes'])
+        self.free = int(data['vagas'])
         self.extra = {
-            'address': data[4],
-            'uid': int(data[2]),
-            'slots': int(data[6]) + int(data[5])
+            'address': data['endereco'],
+            'uid': int(data['id']),
+            'slots': int(data['bikes']) + int(data['bikes'])
```
eskerda: The `slots` field should reflect the absolute number of spaces on an station. It's common to not be there (also adding slots + bikes is not exact too), so you can remove it altogether, since this system does not provide it.

pybikes/ciclosampa.py
```diff
@@ -28,31 +27,30 @@ def update(self, scraper=None):
             scraper = utils.PyBikesScraper()
         scraper.setUserAgent(USERAGENT)
```
eskerda: If the system does not require it, no need to set an special user agent.

pybikes/ciclosampa.py
eskerda: I see the API response also includes a field `status`. Let's add that like
```
extra["status"] = "online" if data['status'] == "Em operação" else "offline"
```
or even, since we do not know which statuses the system has, at least we can assure that we know about "online" like:
```
extra["online"] = data['status'] == "Em operação"
```

Since most API clients support either one or the other.

mmmaia:
@eskerda Thanks for your comments, I tried to address them in the last commit.

pybikes/pegbike.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2014, iomartin <iomartin@iomartin.net>
```
eskerda: Remove this line, or include your own

eskerda:
@mmmaia looks good! See my last comment and this is good to merge, thanks!

pybikes/pegbike.py
mmmaia: done

eskerda:
Perfect @mmmaia , thanks!

PR #367, opened 2019-03-16
Fix bicing system

eskerda:
So much for their "open data" feed. Makes you wonder, how hard it is to
keep one simple endpoint. Citybikes keeps more than 600+


albertvaka:
Looks good to me. The latitude and longitude fields are already floats in this API, so the cast can be removed.

Issue #375, opened 2019-06-07
add .bicyklezadobreskutky, presov, Slovakia


MichalPP:
add presov bike sharing using [open source bikeshare](https://github.com/eskerda/pybikes/blob/master/pybikes/open_source_bike_share.py) with [this url](https://app.bicyklezadobreskutky.sk/command.php?action=map:markers) as source data.

eskerda:
Thanks for reporting the issue and fixing it on a PR @MichalPP 🚲

PR #376, opened 2019-06-07
add prešov bikesharing

MichalPP:
add prešov bikeshareing, fixing #375


eskerda:
Looks good @MichalPP thanks!

PR #380, opened 2019-07-19
Adding support for Clermont-Ferrand C-Vélo new API

lturpinat:
- Modified _smoove.py_ to support new API of C-Vélo at Clermont-Ferrand, FR ;
- also modified _smoove.json_ to make it work : new fetcher and new URL to the API.

Screenshot from the tests :
image

PS : didn't fix the other Smoove bike stations, I only fixed C-Vélo.

### DIFF
--- a/pybikes/data/smoove.json
+++ b/pybikes/data/smoove.json
@@ -22,7 +22,7 @@
                 }
             ]
         },
-        "Smoove":{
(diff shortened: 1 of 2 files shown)


eskerda:
@lturpinat good one, thanks!

PR #394, opened 2019-08-28
update we-cycle url

Dorineli:


eskerda:
Fixed by https://github.com/eskerda/pybikes/pull/400

PR #395, opened 2019-08-28
update edinburgh Just Eat Cycles feed_url

Dorineli:


eskerda:
Good, thanks @Dorineli!

PR #398, opened 2020-01-22
fix bicing ebikes

marcpm:
Since Bicing switched to a new station system, pybike's "has_ebikes"  was always set to False for all stations. Also added number of ebikes.

### DIFF
--- a/pybikes/bicing.py
+++ b/pybikes/bicing.py
@@ -37,6 +37,6 @@ class BicingStation(BikeShareStation):
         self.extra = {
             'uid': int(data['id']),
             'online': data['status'] == 1,
-            'has_ebikes': int(data['electrical_bikes'] > 0),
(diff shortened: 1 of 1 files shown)


eskerda:
Build failed, seems there's a missing `)`. Otherwise, looks good to me

pybikes/bicing.py
```diff
@@ -37,5 +37,6 @@ def __init__(self, data):
         self.extra = {
             'uid': int(data['id']),
             'online': data['status'] == 1,
-            'has_ebikes': 'ELECTRIC' in data['type'],
+            'has_ebikes': int(data['electrical_bikes'] > 0),
```
eskerda:
```suggestion
            'has_ebikes': int(data['electrical_bikes']) > 0,
```

PR #399, opened 2020-01-23
fix bicicoruna url

danielcev:
I'm pretty sure the request to bici coruña failed because now the url is https


eskerda:
Fixes it for me. Thanks @danielcev !

PR #405, opened 2020-02-23
Fix Vcub Bordeaux

trouyer:
Hi,

Here is a fix for the broken v3 Bordeaux parser.

Thomas


eskerda:
Hi @trouyer , thanks!

Issue #384, opened 2019-07-31
Fix velivert


eskerda:
https://www.velivert.fr/#the-map

trouyer:
Hi @eskerda,

open data about stations and real time data exists but doesn't seems to be often updated :
https://www.data.gouv.fr/fr/datasets/donnees-temps-reels-sur-les-veliverts-velo-en-libre-service-de-saint-etienne-metropole/

PR #406, opened 2020-02-27
[gbfs] add velivert- fix eskerda/pybikes#384

trouyer:
Hi @eskerda 

Velivert use gbfs, Hi remove it from smoove.json and add it in gbfs.json.

Thomas


eskerda:
Awesome, thanks!

PR #411, opened 2020-07-13
removed Kyiv Nextbike (Ukraine)

kospl:
https://www.nextbike.ua/en/kyiv/news/nextbike-kyiv-becomes-bikenow/


eskerda:
Just added an issue to look at supporting https://bikenow.com.ua/en/home-2/

PR #413, opened 2020-08-27
Update bike provider for Dresden

farao:
The bike provider in Dresden changed from a cooperation of Nextbike with Sächsische Zeitung (szbike) to a cooperation of Nextbike with DVB (MOBIbike).


eskerda:
@farao thanks!

PR #417, opened 2020-09-21
Add roaming bikes

farao:
This adds (next to stations) the roaming_bikes list to BikeShareSystem and
implements it for Nextbike (seems to be the only currently supported
provider that has roaming bikes that you can park/get outside of
stations).

### DIFF
--- a/pybikes/base.py
+++ b/pybikes/base.py
@@ -16,32 +16,18 @@ class GeneralPurposeEncoder(json.JSONEncoder):
         else:
             return {k: v for k, v in obj.__dict__.items() if not k.startswith('_')}
 
-class BikeShareStation(object):
-    """A base class to name a bike sharing Station. It can be:
-        - Specific (cities):
-            - BicingStation, VelibStation, ...
-        - General (companies):
-            - JCDecauxStation, ClearChannelStation
+class BikeShareInstance(object):
+    """A base class to name a bike sharing instance (either a station or roaming bike).
     """
 
     def __init__(self, name = None, latitude = None, longitude = None,
-                       bikes = None, free = None, extra = None):
+                       bikes = None, extra = None):
         self.name = name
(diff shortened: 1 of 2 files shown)


farao:
Any thoughts on this? :-)

Altonss:
Will this PR be merged?

eUgEntOptIc44:
@Altonss @farao 

please also see #495 it's and nextbike's 'flexzones' here: https://sharedmobility.github.io/Nextbike.html#flexzones (*geojson files*)

For those who read this and wonder what a **'flexzone'** is: it's the area within `roaming_bikes` roam.

Hope @eskerda or possible other maintainer(s) might have the time to look into this any time soon.

theRoboxx:
Is there an update on this request and what is the current problem?
Normally, I use this [website](https://nextbike.net/reservation/?lat=52.3759&lng=9.7320&zoom=15&maponly=1) for flexzone information on devices that don't use the google play services and therefore can't display the map in the nextbike app.

eskerda:
Adding roaming bikes is not a specific nextbike thing, but affects other networks, so we need to consider all implementations when adding this feature to pybikes. Changes on this will also affect the infra where we run pybikes to generate the API, so it's not that trivial. We use lat/lng to generate ids, and roaming bikes change lat/lng constantly, so adding this without planning will backfire. Some networks generate a different id too every time a bike is used for anonymity, so there's that too.

eskerda:
https://github.com/eskerda/pybikes/pull/417

farao:
why was this closed? which commit/pr fixed this?

eskerda:
> why was this closed? which commit/pr fixed this?

Sorry I linked the incorrect PR on the thread: https://github.com/eskerda/pybikes/pull/792

farao:
Thanks :)

PR #419, opened 2020-10-17
Add ebike support to Motivate systems

andyzickler:
Motivate adds a field "num_ebikes_available" to their feed. This checks for that and updates the extra property with the number.

Systems that are supported include bixi-montreal, capital-bikeshare, citi-bike-nyc, cogo, divvy and blue-bikes.

### DIFF
--- a/pybikes/gbfs.py
+++ b/pybikes/gbfs.py
@@ -127,3 +127,7 @@ class GbfsStation(BikeShareStation):
             'returning': info['is_returning'],
             'last_updated': info['last_reported']
         }
+
+        if 'num_ebikes_available' in info:
+            self.extra['has_ebikes'] = True
+            self.extra['ebikes'] = int(info['num_ebikes_available'])
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -41,7 +41,8 @@
                     "Motivate International, Inc.",
                     "PBSC Urban Solutions",
                     "BIXI Montréal"
-                ]
+                ],
+                "ebikes": true
             },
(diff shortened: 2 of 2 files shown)


andyzickler:
@eskerda Any chance I can get this merged? I use Home Assistant to track bikes in my area and knowing if there are ebikes would be very helpful.

eskerda (approved):

pybikes/gbfs.py
```diff
@@ -127,3 +127,7 @@ def __init__(self, info):
             'returning': info['is_returning'],
             'last_updated': info['last_reported']
         }
+
+        if 'num_ebikes_available' in info:
+            self.extra['has_ebikes'] = int(info['num_ebikes_available']) > 0
```
eskerda, line 132: I think this one is always true. Not sure where else has_ebikes is being used right now, but i think it says if the station supports ebikes, no matter the quantity available

eskerda (changes requested):

pybikes/gbfs.py
andyzickler: Thank you for the feedback!  I think you're right. I'm looking at your commit from June https://github.com/eskerda/pybikes/commit/210cb89ab3682517bb67dbf78bc02477835dd4bb and you're setting `has_ebikes` to true if there is an ebike related key in the data from the services. I'll get an update for that shortly.

andyzickler:
@eskerda I've (finally) updated the PR to always set `has_ebikes` to True and I've rebased the branch. I *think* we should be good now, but let me know if there's anything else.

eskerda (approved):

eskerda:
Looking good. Thanks!

PR #422, opened 2021-01-06
Add feed for whole Finland

scop:
This replaces the citybikes-helsinki and foli instances with one covering whole Finland.

The old ones are included in this feed, and are thus redundant now. They are also outdated -- I'll post another alternative PR that tries to update them instead of replacing with a whole Finland feed.


scop:
#423 is the alternative, but as mentioned there, I think I'd prefer this one myself.

PR #427, opened 2021-04-13
velobike: fix this feed provider

ildar:
and add two more sister providers


eskerda:
Good one! 👍 

thanks!!

ildar:
🙂
When it'll get to the API server?

eskerda:
Should be there by now 🚀

ildar:
It is! thanks a lot!

PR #429, opened 2021-07-20
Fix bicincitta.py url param

MarStefo:
The remote server was "upgraded" to https and the original http address route to the homepage.

Example of old url for Udine (Italy) station: 
http://www.bicincitta.com/frmLeStazioniComune.aspx?ID=126 ==> it is routed to homepage (wrongly)
https://www.bicincitta.com/frmLeStazioniComune.aspx?ID=126 ==> OK.

The systems engineers have done a great job 👏...


eskerda:
LGTM, thanks a lot!

Issue #432, opened 2021-09-02
Vel'In from the city of Calais (Fr) is not up to date


Altonss:
The last information was 421 days ago, there seems to be an issue...

Altonss:
Just found a feed where the data comes from on the official site (https://www.vel-in.fr/EnTempsReel):
https://velo-vision-prod.azurewebsites.net/calais/oybike/stands.nsf/getsite?site=calais&format=json

Just need to implement it ^^

PR #440, opened 2021-10-10
Fix Vél'In feed in Calais(FR)

Altonss:
Update feed for Vél'In in Calais(FR)
This fixes issue #432 
Ping @eskerda , I hope I have done this right and that it can get merged ;)

P.S. At the end of the new URL feed I scraped from the official Vél'In website (https://www.vel-in.fr/EnTempsReel) there is key=veolia. Can this remain or is there an issue?

### DIFF
--- a/pybikes/data/veloway.json
+++ b/pybikes/data/veloway.json
@@ -34,7 +34,7 @@
                         "longitude": 1.85,
                         "country": "FR"
                     },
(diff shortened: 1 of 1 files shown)


eskerda:
Thanks! that works. No problem on having a hardcoded parameter on the url.

I see no other systems on veloway work, so I will try to fix or remove them.

PR #448, opened 2021-10-27
Add Ecovelo provider and associated cities

Altonss:
Hello @eskerda,
I managed to find a new provider and make it work with a lot of cities. But I have 3 questions:
- For some cities with only mechanical bikes, I habe the label 'ebike' set to true in meta, because I have set this as a general parameter for al ecovelo providers (because most have ebikes). Is this a problem?
- What means 'has_ebike' for a station? For the moment all stations have it, because the api always displays it.
- Should we also consider electric scooters? See the code to know more, but in the api there is an easy way to find how much electric scooter there are in a station.
Thanks in advance for your feedback ;)


Altonss:
Any update concerning this PR @eskerda ?

eskerda:
Hello @Altonss! This system really has a lot of meat, so it took me a while to get some time to review. Regarding your questions:

> For some cities with only mechanical bikes, I habe the label 'ebike' set to true in meta, because I have set this as a general parameter for al ecovelo providers (because most have ebikes). Is this a problem?

On that case, we can add it on the `json` data file and not on the class wide object.

> What means 'has_ebike' for a station? For the moment all stations have it, because the api always displays it.

We added this field before `ebikes`, since at the time we didn't consider we would ever have counts. Some apps started depending on it, and so, now we have both `ebikes` and `has_ebike`. Ideally, we will remove `has_ebike` field on the future, since it is redundant with the number of ebikes.

> Should we also consider electric scooters? See the code to know more, but in the api there is an easy way to find how much electric scooter there are in a station.

Usually I would have said no. But I think yes! My general opinion is that this project is only for bicycles, but the future looks a bit more complicated than that.. since this system in particular coexists bicycles and scooters, I see no reason to not put it on.

This PR has been open for a while, so on the interest of getting this system ASAP I will merge it as it is, and if you do not mind I will add the changes once it lands on master.

Amazing work!

Altonss:
Thanks a lot for the merge of this PR! I noticed that the data on the citybik.es API is now 4 days old for the Ecovelo provider. Is this normal or maybe do you know the issue?

eskerda:
Meh...

```
You are not authorized to make this call. Please use our GBFS if you need open data at https://api.gbfs.v1.0.ecovelo.mobi.
(e.g. https://api.gbfs.v1.0.ecovelo.mobi/gbfs/{program}/{language}/station_status with your city for {program} and "fr" for {language}.)
```

Did you know about the GBFS feeds?

Altonss:
No I did not know about it at all :/ I didn't found any documentation about that... Well then all this work was useless lol :smile:

eskerda:
I did not find any documentation either, so that makes two of us :) All work including extra was good work. There's a gbfs feed, but still, it contains less information.  That's my gripe with gbfs, it will miss relevant information compared what you got here. In terms of total freedom, you could never write an app that would compete on equal terms with their own. 

For now setting a browser user agent is enough. Let's move these to the GBFS feed at some point

Altonss:
> I did not find any documentation either, so that makes two of us :) All work including extra was good work. There's a gbfs feed, but still, it contains less information. That's my gripe with gbfs, it will miss relevant information compared what you got here. In terms of total freedom, you could never write an app that would compete on equal terms with their own. But honestly, at this point it's a battle I am tired of having to fight.

I totally agree, they retain some precious information by having a very limited gbfs feed :/


> For now setting a browser user agent is enough. Let's move these to the GBFS feed at some point

Well if that is working great there would not really be a need to change to GBFS. But you are right, as pybikes is only providing a "general" api, the very detailed infos are not really needed. But if it works as is with the fix, that's great :)

eskerda:
They added "PyBikes" to their list of unauthorized user agents, but my guess is that if we keep using it they will add any other means of authentication soon. Once this starts it's mostly a cat and mouse game.

PR #455, opened 2021-12-24
fix(veloway) python 3 compatibility

eskerda:


Altonss:
That's great for the python 3 compatiility, thank you!
Do you know what still needs to be done do definitely switch to Python 3?

PR #465, opened 2022-03-14
Add some old fields to velib

eskerda:
It seems to me https://velib-metropole-opendata.smoove.pro/opendata/Velib_Metropole/station_status.json has some fields that are not on the GBFS specification. And that's fine. Either that, or it was part of an old GBFS version?

This makes GBFS accept a `station_cls` field.. I do not want to call this a factory, but it is indeed a factory. Seemed to me the most straightforward way of adding these fields to velib without messing with GBFS and other instances.

(re)adds:
- banking
- uid as station code
- ebikes

### DIFF
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -963,21 +963,6 @@
                 "longitude": -2.7603
             },
             "feed_url": "https://vannes-gbfs.klervi.net/gbfs/gbfs.json"
-        },
-        {
-             "tag": "velib",
-             "meta": {
-                 "latitude": 48.856614,
-                 "country": "FR",
-                 "name": "Velib' Métropôle",
-                 "longitude": 2.3522219,
-                 "city": "Paris",
-                 "license": {
-                     "name": "OPEN LICENCE 2.0",
-                     "url": "etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf"
-                }
-             },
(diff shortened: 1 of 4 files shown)


Altonss:
A lot of GBFS feeds have ebikes! And those feeds are often not yet used with the full potential yet.
The same goes for adresses: for a lot of cities with gbfs, the address field is set to "null" instead of just not having one. This confuses some applications using the citybik.es api

eskerda:
> A lot of GBFS feeds have ebikes! And those feeds are often not yet used with the full potential yet.

 If you look closely at the velib feed, the ebikes are implemented on a way that is not on the GBFS specification

Altonss:
Oh that might be the case I don't know the specifications enough. What I mean is that we need to manually add the ebike capability to a lot of cities like for this still open PR https://github.com/eskerda/pybikes/pull/461 instead of having an automatic ebike detection :smile:

eskerda:
> Oh that might be the case I don't know the specifications enough. What I mean is that we need to manually add the ebike capability to a lot of cities like for this still open PR https://github.com/eskerda/pybikes/pull/461 instead of having an automatic ebike detection 😄

I keep seeing we are adding this to feeds, but this is not being used anywhere on the API. Why are we doing it?

Altonss:
> I keep seeing we are adding this to feeds, but this is not being used anywhere on the API. Why are we doing it?

As of know, the ebikes for #461 are not displayed on citybik.es so that's why they are adding it.

eskerda:
I guess I forgot to add it to the API

PR #498, opened 2022-08-13
Added support for VRN Nextbike in Landau (Pfalz)

vosskuhle:
Extracted data from <https://maps.nextbike.net/maps/nextbike-live.json?city=678>


eskerda:
👍 thanks!

Issue #471, opened 2022-04-14
Velobike_ru seems broken


Altonss:

eskerda:
At the moment velobike is rejecting requests that do not come from a browser, by issuing a javascript challenge that generates a nonce that later returns a valid session id that must be included on the request. It's using anti ddos protection from qrator labs.

Only solution on this is either checking how long it takes for a valid session to expire and see if we can reasonably use that, or add some extra dependencies on the project to communicate with a js engine and solve these kinds of js challenges. So far this is the only one so I am inclining towards no

eskerda:
Dupe https://github.com/eskerda/pybikes/issues/839

PR #503, opened 2022-09-10
Update velobike_ru

Manu1400:
Some updated about config file:
add 2 feeds
* an university
* Tyumen city

remove an old city (Nizhniy Novgorod)

simplify an URL:
- [x] check: same data content without the param


Altonss:
This is related to #471 
The feed for Moscow seems broken for example... at least not updating often enough.

PR #518, opened 2022-12-20
Clermont-Ferrand: Update to v2 GBFS

Manu1400:
documentation: https://transport.data.gouv.fr/datasets/etat-des-stations-de-velos-en-libre-service-c-velo-gbfs-du-syndicat-mixte-des-transports-en-commun-de-lagglomeration-clermontoise-smtc-ac-1


Manu1400:
not a GBFS feed, my mistake

PR #519, opened 2022-12-20
Fix Miami Beach, FL feed

Manu1400:
source: http://www.citibikemiami.com/station-map


eUgEntOptIc44 (changes requested):
Hi @Manu1400 

It might be better to use the `https` scheme. (also see #522, #523)
You'll probably intended to use 'https' but typed 'http'. At least this happens to me on a regular basis.

eskerda:
👍 Thanks!

PR #522, opened 2022-12-21
[improvement] WhiteBikes: move to HTTPS

Manu1400:
prefer https://whitebikes.info/command.php?action=map:markers

## information
http://whitebikes.info/command.php?action=map:markers redirect to https


eskerda:
👍 thanks!

PR #523, opened 2022-12-21
BikeSantos: prefer HTTPS

Manu1400:
## goal
for security, prefer HTTPS


eskerda:
I am fine with the changes, but would like to know more about the reasoning. There's nothing intrinsically bad about using regular http for non privacy sensitive requests. My rule of the thumb for this is for redirects from http to https, causing extra requests.

PR #526, opened 2022-12-21
[minor change] fix a typo in name "Vélib' Métropole"

Manu1400:
source: https://fr.wikipedia.org/wiki/V%C3%A9lib'_M%C3%A9tropole


eskerda:
👍 many thanks @Manu1400

PR #527, opened 2022-12-21
Fix markdown links in CI jobs

Manu1400:
Add prefix "user-content-"

## documentation
i.e: "cls-keolisstar-3.9.16" => "user-content-cls-keolisstar-3.9.16"


eskerda:
Still does not work for me @Manu1400 , I think there's something janky with the summary links and anchors. Did you manage to make it work?

eskerda:
Fixed in https://github.com/eskerda/pybikes/pull/560

PR #517, opened 2022-12-20
Fix Velopop (Avignon)

Manu1400:
documentation: https://transport.data.gouv.fr/datasets/velos-libre-service-grand-avignon-velopop-disponibilite-en-temps-reel

PR #535, opened 2023-01-05
WIP: check if coordinates of GBFS stations are None

eUgEntOptIc44:
As mentioned in #517 I had noticed that the [gbfs.py](https://github.com/eskerda/pybikes/blob/24e40351678fa22f74af0aa7da3f7a326eee7e2e/pybikes/gbfs.py#L114) parser fails on feeds containing stations with missing (`null` / `None`) coordinates.

This PR is meant to serve as a starter for a fix of this inconvience. As of writing the parser would silently skip stations with missing coordinates. As noted inline I think that some feedback to the user would be required. Should the parser `print()` an error message? Throwing an **exception** would to my personal understanding break the parsing loop.

As the title suggests this PR is WIP as of publishing. Please feel free to comment on this and provide feedback.

### DIFF
--- a/pybikes/gbfs.py
+++ b/pybikes/gbfs.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2010-2012, eskerda <eskerda@gmail.com>
+# Copyright (C) 2010-2023, eskerda <eskerda@gmail.com>
+# Copyright (C) 2023, eUgEntOptIc44 (https://github.com/eUgEntOptIc44)
 # Distributed under the AGPL license, see LICENSE.txt
 
 import json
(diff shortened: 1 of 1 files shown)


eskerda:
skipping stations with missing lat/lng should only happen on a per-network basis and not on all the GBFS networks. The solution I had in mind was to subclass Gbfs for this particular network and add any needed quirks there, similar to https://github.com/eskerda/pybikes/blob/afe31556a1b4564eb0394dc13fd74d0dcd127a4d/pybikes/velib.py#L1-L33

eskerda:
An alternative solution would be to add a flag for it, like `skip_invalid = true` on class instantiation or something, but not as a general rule because that can potentially hide issues in all other gbfs networks

PR #540, opened 2023-02-01
minor improvement to test_data

andrehora:
This PR adds two tests to `test_data` to cover some untested exception cases, ie, when the tag is empty.

Thus, it creates two tests:
- test_find_not_found_empty_tag
- test_get_not_found_empty_tag


eskerda:
Hey @andrehora, thanks for your contribution! Unfortunately, it does not add substantial enough coverage to be a relevant test.

Closing it for now. Let me know if you think there are particular cases that are not covered enough on the data tests and we can reopen it.

PR #552, opened 2023-06-15
BikeshareIE: added Waterford

futuretap:
renamed the system to "TFI Bikes"


eskerda:
Looking good, thanks @futuretap!

PR #558, opened 2023-07-05
cyclopolis: fix free bikes

futuretap:
I noticed that the number of available bikes wasn't parsed correctly


eskerda:
I see that none of the `google` systems are online, so we cannot check if these changes work there. I am ok with adding these changes for now and moving on a separate parser if anything changes. Although I suspect most systems will migrate to mapbox. Thanks @futuretap!

futuretap:
Yeah, coverage is ridiculous. Looks like they’re struggling. Let’s hope we can help them get more traction.

PR #562, opened 2023-07-07
New provider in Saint-Etienne, same name as before but new GBFS stream link

Ephtolens:
I've also corrected the accent in city and name of object, is it ok ?


eskerda:
Yes it is. Thanks!

PR #564, opened 2023-07-11
Add rental URIS info in GBFS station extra data.

Ephtolens:
Some GBFS includes deeplinks to Android/iOS app, could be great to retrieve them in extra data of GBFS stations ?
Based on these examples : https://github.com/MobilityData/gbfs/blob/v2.3/gbfs.md#examples


eskerda:
The more the merrier, thanks!

PR #574, opened 2023-07-26
Use dedicated availability feed hostname

j0kan:
In the meantime we created a dedicated host just to fetch our availability feeds. Use this to improve latency, performance and reliance.


eskerda:
Thanks @j0kan! Happy to discuss if there's anything else we must do to integrate better with nextbike

PR #579, opened 2023-08-02
Added Bicincitta schemes

futuretap:
There are also a number of IDs used for Weelo schemes (e.g. 277, 281, 282, 287, 288, 293, 294), however they return different (outdated?) data than the Weelo app, so I didn't include them. Would be interesting, if they work using `WeeloAPI`. Can you check?

### DIFF
--- a/pybikes/data/bicincitta.json
+++ b/pybikes/data/bicincitta.json
@@ -1525,6 +1525,58 @@
             },
             "tag": "bicimia",
             "endpoint": "https://bicimia.bresciamobilita.it"
+        },
+        {
+              "city_ids": [
+                  268
+              ],
+              "meta": {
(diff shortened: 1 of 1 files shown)


eskerda:
> There are also a number of IDs used for Weelo schemes (e.g. 277, 281, 282, 287, 288, 293, 294), however they return different (outdated?) data than the Weelo app, so I didn't include them. Would be interesting, if they work using `WeeloAPI`. Can you check?

I do not understand, where are these ids coming from?

futuretap:
Just tested them. :)

eskerda:
We are going to migrate cities to Weelo implementation, but I want to do it conservatively, we have a shit ton of bicincitta services added, and most of them are completely useless

PR #589, opened 2023-08-05
Update Bicimad feed url

MatiasConTilde:


eskerda:
Is there something wrong with the current feed?

pybikes/bicimad.py
```diff
@@ -7,10 +7,15 @@
 import json
 
 from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
+from pybikes.contrib import TSTCache
 
 
+PASSKEY = 'C3D0E659D8D397782B414AB6FCC477B5C727435FE91E72069D34CEBB2C1491B3B8563FDAC043EA704660C0E87E6FE503C39D38FF43F7447563B1E437B349ACEC'
+CLIENTID = '8ff527f9-f85b-45ef-b1b2-bd9eb59e0fff'
```
eskerda, line 14: Let's not commit these secrets

MatiasConTilde:
It has missing a lot of new stations

pybikes/bicimad.py
MatiasConTilde: I implemented it more like https://github.com/eskerda/pybikes/blob/master/pybikes/ecobici_ba.py, which seemed simpler and more fitting for this

MatiasConTilde: The "leak" is not a security issue as those are the default public credentials for the app. Also, my goal with this PR is that the citibik.es API gets updated, how can I inform that those are valid credentials so they can use them in prod?

eskerda: It is not a security issue, but they could fill a take down request on GitHub if they wanted. Not committing secrets, even when these are public, reduces our legal surface

eskerda: > The "leak" is not a security issue as those are the default public credentials for the app. Also, my goal with this PR is that the citibik.es API gets updated, how can I inform that those are valid credentials so they can use them in prod?

You can send them by email

eskerda:
Looking good. Thanks!

pybikes/bicimad.py
```diff
@@ -7,43 +7,64 @@
...
+    def __init__(self, tag, meta, feed_url, key):
         super(Bicimad, self).__init__(tag, meta)
         self.feed_url = feed_url
+        self.key = key
+
+    @staticmethod
+    def authorize(scraper, key):
+        request = scraper.request
+
+        accesstoken_scraper = PyBikesScraper(cache)
```
eskerda, line 35: I like this idea for keeping the token valid for a period of time :+1: thanks!

PR #590, opened 2023-08-10
velobike_v2: the modified version of velobike

ildar:
fixes velobike.kz provider

### DIFF
--- a/pybikes/velobike.py
+++ b/pybikes/velobike.py
@@ -25,6 +25,9 @@ class Velobike(BikeShareSystem):
         stations = []
 
         data = json.loads(scraper.request(self.feed_url))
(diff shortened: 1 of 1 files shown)


eskerda (changes requested):
Hey @ildar, thanks for your contribution!

As far as I see, the only difference between `velobike` and `velobike_v2` is that the information is inside `data` instead of directly on the array response. I think just for this one, the cleanest would be to not duplicate the entire implementation and instead add a hack on `velobike` implementation with a comment, like:


```
# Some feeds have it inside a list, others within `data`
if isinstance(data, dict):
    data = data['data']
```

ildar:
good idea! thanks, I'll do that.

eskerda:
Great! Thanks @ildar

PR #605, opened 2023-09-18
Add NextBike Prerov, CZ

Golbinex:
This year NextBike started in Prerov, Czech Republic.
Values were extracted from official NextBike API.

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -2672,6 +2672,18 @@
                 "latitude": 52.2581,
                 "longitude": 20.9949
             }
+        },
+        {
+            "domain": "nr",
+            "tag": "nextbike-prerov",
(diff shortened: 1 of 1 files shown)


Golbinex:
Sorry, I thought that domain equals with country code, because neighboring city, Prostejov, coincidentally has "cz" domain, which is same like our country code 😁

Now the domain is corrected and tested.

eskerda:
image

eskerda:
Looking good, thanks @Golbinex!! 

> @eskerda Sorry, I thought that the domain equals with country code, because neighboring city, Prostejov, coincidentally has "cz" domain, which is same like our country code 😁
> 
> Now the domain is corrected and tested.

Nextbike domains are a bit discretionary, so it happens a lot

Looking good, thanks @Golbinex !!

Issue #286, opened 2017-10-11
Add Tokyo's "Docomo Cycle"


nicolas-raoul:
URL: http://docomo-cycle.jp/tokyo-project/
This URL contains a map that shows bike stations, so I guess information can be extracted from that.

The bicycles are separated in "communities" such as "minato" or "chiyoda", I am not sure whether they should be considered different systems or not.
A dozen other cities use the same provider, presumably powered by the same technology. I added all to https://github.com/eskerda/pybikes/wiki/Missing-bike-systems
I can translate from Japanese whatever you want, just ask me.

There used to be an Android app that showed how many bikes are available at each station, but it seems to have been retired, now it seems that users are requested to use the website.

Thanks a lot for maintaining this project! :-)

reustle:
Hey, could you please clarify which app you are talking about? I'm guessing it is only available in the Japanese app store?

reustle:
Okay, you said "Android app" so I wanted to clarify since I haven't found any native apps. Thanks!

nicolas-raoul:
@reustle oops I did not remember that, I edited the issue :-) Thanks!

nicolas-raoul:
On the opposite, I can confirm that the bicyles are still present and used by many :-)

markszabo:
I've recently moved to Tokyo and using Docomo Cycle ever since. The website has limited information (only an embedded Google Maps with the stations and no information on the available bikes per station), but their [android app](https://play.google.com/store/apps/details?id=com.d_bikeshare)(only available from Japan I guess) has the data on the individual bikes per station. I'll keep looking into the communication of the app to figure out their API. I'll keep you posted here.

reustle:
I currently host station data on https://TokyoBikeShare.com (which is currently not running, back online soon). looking forward to hearing more about what you find on their API :)

markszabo:
This is what I got sofar: https://repl.it/repls/AllEssentialBrackets

The output can be copied into an HTML file to draw the stations to a map: https://jsfiddle.net/r1pxda2m/

(Files are also attached for later reference should the above URLs die: [docomo1.zip](https://github.com/eskerda/pybikes/files/3554805/docomo1.zip))

The colors on the map:
- green: this is a station and I could get the number of bikes there (even if it was 0) - there are 1360 stations like this
- red: not a station. Has a `tsc_cd` id, but returned something else than 200 - there are 64 things like this 
- gray: not a station, doesn't have a `tsc_cd` - there are 25 things like this

[...]

PR #606, opened 2023-10-02
Add Docomo Tokyo Bike Share

martgnz:
GBFS feed comes from ODPT, a Japanese public transport platform: https://ckan.odpt.org/dataset/c_bikeshare_gbfs-d-bikeshare.

Closes https://github.com/eskerda/pybikes/issues/286.


eskerda:
Looking good! Thanks 🚀

PR #607, opened 2023-10-03
Add Hello Cycling

martgnz:
This is a nation-wide bike-sharing network in Japan that has multiple providers. Data comes from the same source as https://github.com/eskerda/pybikes/pull/606.


eskerda:
Merged in https://github.com/eskerda/pybikes/commit/888cc28aaceba87eabba7e0aaa39c332d87ae505

Thanks @martgnz !!!

PR #614, opened 2023-10-05
Add Biki

martgnz:
Adds a new Bicing system in Valladolid: https://biki-valladolid.es/es

https://www.elnortedecastilla.es/valladolid/como-utilizar-biki-20230206110850-nt.html


eskerda:
Awesome! thanks

PR #616, opened 2023-10-05
Add Wegoshare

martgnz:
Add a new system with a few active bikeshares in Portugal and in the US: https://wegoshare.com/explore-our-network/

https://algira.almeirim.pt
https://rmbikes.cm-riomaior.pt
https://figas.cm-figfoz.pt/
https://summitbikeshare.com/
https://rvabikes.com/map

### DIFF
--- a/pybikes/data/wegoshare.json
+++ b/pybikes/data/wegoshare.json
@@ -58,6 +58,21 @@
                 ]
             },
             "tag": "summit-bike-share"
+        },
+        {
+            "endpoint": "https://api.rvabikes.com",
+            "meta": {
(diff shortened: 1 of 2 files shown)


eskerda:
Not sure about `totalLockedCycleCount` vs `primaryLockedCycleCount` The map seems to display `totalLocked..` but then when you click on the station the actual info that comes out seems to be consistent with `primaryLocked..`

eskerda:
> RVA Bike Share in Richmond, VA seems that it should be open but all stations look closed on the map

Seems open now 

image

PR #617, opened 2023-10-05
Add new Tembici systems in LATAM

martgnz:
Add a few new systems managed by Tembici, including Bogotá and Buenos Aires (replacing the custom scraper for a GBFS entry).

~For some reason rivibike is failing the tests. Not sure why!~ The system was already in, using an old system that failed. I have removed it and the new GBFS endpoint is working.


eskerda:
Great, thanks!

PR #618, opened 2023-10-06
Add Citybike Lima

martgnz:
A system in Lima's district of Miraflores. It is a [partnership](https://www.facebook.com/SmooveSAS/posts/2457798104276208/) between Smoove and Moventia. Not sure if it is a 100% Velobike system, but the property names are very close.

https://www.citybikelima.com/es

### DIFF
--- a/pybikes/data/smoove.json
+++ b/pybikes/data/smoove.json
@@ -15,6 +15,27 @@
                     "feed_url":"http://cli-velo-batumi.gir.fr/sag_map.html"
                 }
             ]
+        },
+        "CityBikeLima": {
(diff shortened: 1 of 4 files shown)


eskerda:
Even if property names are very close I do not think it is worth supporting the two kinds of feeds on the same class. Let's make this its own snowflake module, if you want inside `smoove.py` and `smove.json`.

Now that I remember, we used to have this kind of feed supported, see this commit https://github.com/eskerda/pybikes/commit/ec019cf3c946e3ddfd1c1ada2ef7a61e1387e211

PR #620, opened 2023-10-06
Add Bicielx

martgnz:
Add this system in Elx: https://bicielx.es/

Have fun! https://gestion.bicielx.es/mapa_prestamo.php


eskerda:
This feed is incredible

PR #621, opened 2023-10-07
Add Skrova Island Bike

martgnz:
A service with one station in Skrova, an island in the Lofoten archipelago nearly 70º degrees north: https://skrovabysykkel.no/en

It is an Urban Sharing service but unfortunately it is missing from the normal GBFS feeds.


eskerda:
New northernmost working bike sharing system nice

PR #622, opened 2023-10-07
Add Bicisanvi and dbizi

martgnz:
Add Bicisanvi, a system in Sant Vicent del Raspeig (ES): https://www.bicisanvi.es/online.php

Add dbizi, a system in San Sebastián (ES): https://www.dbizi.eus/es/plano-y-estaciones

It uses the same Mibisivalencia scraper, which I renamed to Movus. It's the company behind both systems: https://movus.es/

San Sebastián has electric bikes so the logic has to change a bit.

### DIFF
--- a/pybikes/movus.py
+++ b/pybikes/movus.py
@@ -7,21 +7,22 @@ import re
 
 from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
 
-BASE_URL = 'https://www.mibisivalencia.es/mapa/mapa.php'
 
-
-class MiBisiValencia(BikeShareSystem):
+class Movus(BikeShareSystem):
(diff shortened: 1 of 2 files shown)


pybikes/movus.py
```diff
@@ -0,0 +1,72 @@
...
+class Movus(BikeShareSystem):
+    sync = True
+
+    company = ['Movilidad Urbana Sostenible SLU']
+
+    def __init__(self, tag, meta, feed_url, ebikes):
+        if 'company' in meta:
+            meta['company'] += Movus.company
+        else:
+            meta['company'] = Movus.company
```
eskerda, line 20: Not needed
```suggestion

```

Issue #233, opened 2017-02-07
Add Merano bike share


eskerda:
http://bike.meran.eu/

eskerda:
Seems like feeds might be

Stations and other info are documented here: http://geodata.integreen-life.bz.it/geoserver/web/
Precisely: http://geodata.integreen-life.bz.it/geoserver/edi/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=edi:Bikesharing&maxFeatures=50&outputFormat=json

This endpoint also offers information: http://ipchannels.integreen-life.bz.it/doc/
ie: http://ipchannels.integreen-life.bz.it/bikesharingFrontEnd/rest/get-station-details

eskerda:
https://otp.opendatahub.com/otp/routers/openmove/bike_rental

martgnz:
This feed is returning weird numbers for Merano: 0, 100 or 1000 bikes available. If the endpoint starts to return the right numbers it will be easy to add once https://github.com/eskerda/pybikes/pull/623 is in.

PR #623, opened 2023-10-07
Add Bolzano

martgnz:
For now this only has one network in Bolzano, IT. The JSON holds more data: Merano (https://github.com/eskerda/pybikes/issues/233) and Papin. Unfortunately both the Merano and Papin networks are returning bogus numbers with only 0, 1, 100 or 1000 bikes available.

I'm not sure if Papin is actually a bike-sharing service. It looks like a rental bike company with stations in multiple towns.

### DIFF
--- a/pybikes/data/openmove.json
+++ b/pybikes/data/openmove.json
@@ -1,0 +1,20 @@
+{
+    "instances": [
+        {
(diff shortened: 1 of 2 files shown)


eskerda:
works, thanks!!!! 🚀

PR #626, opened 2023-10-13
[gbfs] add configurable cache deltas for feeds

eskerda:
Current gbfs implementation does 4 requests per update on a worst case scenario. Gbfs is as it is by design, but we can try to make it a bit easier on the servers by adding a particular delta on specific parts of the feed that do not change that often, like station_information, vehicle_types or the main gbfs index response.


eskerda:
Closed in favour of https://github.com/eskerda/pybikes/pull/642

Issue #584, opened 2023-08-03
Add stavanger kolumbus system


eskerda:
See https://github.com/eskerda/pybikes/issues/499#issuecomment-1358678219

PR #633, opened 2023-10-18
Add Kolumbus - Rogaland region

martgnz:
This system actually covers Stavanger and the whole region of Rogaland: https://www.kolumbus.no/en/travel/bicycle/bysykkelen/

Please take a look at the logic and let me know if it looks right.

Closes https://github.com/eskerda/pybikes/issues/584.

### DIFF
--- a/pybikes/kolumbus.py
+++ b/pybikes/kolumbus.py
@@ -1,0 +1,59 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2023, Martín González Gómez <m@martingonzalez.net>
+# Distributed under the LGPL license, see LICENSE.txt
+
+import json
+
+from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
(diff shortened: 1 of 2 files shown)


pybikes/kolumbus.py
```diff
@@ -0,0 +1,59 @@
...
+        bikes = int(data['available'])
+        slots = int(data['capacity'])
+
+        self.bikes = bikes
+        # there are only a few charging slots available per station
+        # but people can leave bikes close if there is no room
+        # this means we could have more bikes than slots: in that case count it as 0
+        self.free = max(slots - bikes, 0)
+        self.extra = {
+            'ebikes': True,
```
eskerda, line 56: `ebikes` is not a boolean but the (int) number of available e-bikes when there are both normal bikes and e-bikes.

martgnz: thank you. all bikes in this system look like ebikes

PR #641, opened 2023-10-27
Add fifteen service

Altonss:
I'm not really sure for the city field though...


eskerda:
Hey @Altonss, thanks!

Regarding the city name, I would not worry. It's not a city, but it's descriptive.

PR #647, opened 2023-11-29
fix(velib): change gbfs url

Ephtolens:
The GBFS url for Velib (Paris) has changed since 26 September 2023 (see [Velib website](https://www.velib-metropole.fr/donnees-open-data-gbfs-du-service-velib-metropole))
Here is the new one -> [new gbfs url](https://velib-metropole-opendata.smovengo.cloud/opendata/Velib_Metropole/gbfs.json)

### DIFF
--- a/pybikes/data/velib.json
+++ b/pybikes/data/velib.json
@@ -14,7 +14,7 @@
                 },
                 "ebikes": true
             },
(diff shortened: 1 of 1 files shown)


eskerda:
👍  thanks!

PR #653, opened 2023-12-17
Add Careem Bike

martgnz:
GBFS feed for Dubai, Abu Dhabi and Medina.


eskerda:
WDYT about creating a `careem` class that inherits from GBFS and is listed as `unifeed`

martgnz:
Much better, done!

PR #662, opened 2023-12-26
Add QroBici

martgnz:
A system in Querétaro, México: https://municipiodequeretaro.gob.mx/qrobici/

### DIFF
--- a/pybikes/utils.py
+++ b/pybikes/utils.py
@@ -41,7 +41,6 @@ def sp_capwords(word):
 class PyBikesScraper(object):
     proxy_enabled = False
     last_request = None
-    ssl_verification = True
     requests_timeout = 300
     retry = False
     retry_opts = {}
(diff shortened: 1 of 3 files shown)


pybikes/utils.py
```diff
@@ -41,7 +41,6 @@ def sp_capwords(word):
 class PyBikesScraper(object):
     proxy_enabled = False
     last_request = None
-    ssl_verification = True
```
martgnz, line 44: The system didn't work locally without disabling ssl verification. I included it in the request call, please let me know if this should be done differently.

eskerda: You can set this without changing the class specification

```
foo = PyBikesScraper()
foo.ssl_verification = False
```

eskerda: But having it as an argument works for me too

PR #678, opened 2023-12-30
Fix Clujbike

martgnz:
Endpoint shut down. Loads data from this map instead: https://portal.clujbike.eu/harta-statii-biciclete

### DIFF
--- a/pybikes/clujbike.py
+++ b/pybikes/clujbike.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
+# Copyright (C) 2023, Martín González Gómez <m@martingonzalez.net>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import re
 import json
 
-from .base import BikeShareSystem, BikeShareStation
-from . import utils
+from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
 
 
 class Clujbike(BikeShareSystem):
-
-    sync = True
 
     meta = {
         'system': 'Clujbike',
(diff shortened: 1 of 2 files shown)


pybikes/clujbike.py
```diff
@@ -19,70 +20,34 @@ def __init__(self, tag, feed_url, meta):
...
-            'IdStatus' : '0',
-            'Status' : '',
-            'StatusType' : '',
-            'Latitude' : '0',
-            'Longitude' : '0',
-            'IsValid' : 'true',
-            'CustomIsValid' : 'false',
-            'Id' : '0',
-        }
+        data = re.search(r'var stations = (.*?;)', raw).group(1).replace('}];', '}]')
```
eskerda, line 27:
```suggestion
        data =  re.search(r'var stations = (\[.*?\]);' raw)
```

martgnz: thank you!!!

pybikes/clujbike.py
```diff
@@ -19,70 +20,34 @@ def __init__(self, tag, feed_url, meta):
...
+        for station in json.loads(data):
+            stations.append(ClujBikeStation(station))
 
         self.stations = stations
+
+
+class ClujBikeStation(BikeShareStation):
+    def __init__(self, data):
+        super(ClujBikeStation, self).__init__()
+        print(data)
```
eskerda, line 40:
```suggestion
```

martgnz: sorry, very embarrasing

pybikes/clujbike.py
```diff
@@ -19,70 +20,33 @@ def __init__(self, tag, feed_url, meta):
...
+
+        self.name = data['StationName']
+        self.latitude = float(data['Latitude'])
+        self.longitude = float(data['Longitude'])
+
+        self.bikes = int(data['OcuppiedSpots'])
+        self.free = int(data['EmptyDoors'])
+        self.status = 'offline' if data['Status'] == 'Offline' else 'online'
+        self.extra = {
+            'slots': self.bikes + self.free,
```
eskerda, line 49: Not sure we can trust to properly represent the number of slots that stations have. I see there are many stations with 19 slots. I would go with removing it unless it's constantly accurate.

martgnz: Sure, just thought it could be helpful. Removed!

PR #693, opened 2024-01-01
Add Aksu, Aral and Kuitun

martgnz:
http://www.akspub.com/html/bikezt/map/
http://www.kuitunbike.com/html/ktmap/

### DIFF
--- a/pybikes/aksu.py
+++ b/pybikes/aksu.py
@@ -22,7 +22,7 @@ class Aksu(BikeShareSystem):
 
         raw = scraper.request(self.endpoint)
 
-        marker_var = re.search(r'var ibike = (.*)', raw, re.DOTALL)
+        marker_var = re.search(r'var ibike = (.*)', raw)
         markers = json.loads(marker_var.group(1))
 
         stations = []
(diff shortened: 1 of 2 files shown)


pybikes/aksu.py
```diff
@@ -0,0 +1,52 @@
...
+        super(Aksu, self).__init__(tag, meta)
+        self.endpoint = endpoint
+        self.bbox = bbox
+
+    def update(self, scraper=None):
+        scraper = scraper or PyBikesScraper()
+
+        raw = scraper.request(self.endpoint)
+
+        marker_var = re.search(r'var ibike = (.*)', raw, re.DOTALL)
```
eskerda, line 25:
```suggestion
        marker_var = re.search(r'var ibike = (.*)', raw)
```

pybikes/aksu.py
```diff
@@ -0,0 +1,52 @@
...
+class AksuStation(BikeShareStation):
+    def __init__(self, data):
+        super(AksuStation, self).__init__()
+
+        self.name = data["name"]
+        self.latitude = float(data["lat"])
+        self.longitude = float(data["lng"])
+
+        self.bikes = int(data["availBike"])
+        self.free = int(data["capacity"])
```
eskerda, line 47: This field looks like the total capacity

PR #697, opened 2024-01-02
Automatic bbox filtering

eskerda:
This PR adds an utility class that magically filters by a bbox or bounds argument.

We have many systems that use the same boilerplate for filterting bounds. Not all cases are the same, for instance on gbfs and nextbike we prefilter before parsing the stations. For the general usecase, this utility class is the best of both worlds.

I have decided against adding it to the base class itself because of unexpected behaviors using properties. We could set a flag over at the class, like `disable_bbox_filtering`, but having that flag (undocumented) would probably be less explicit than using an utility class.

Let me know what you think

### DIFF
--- a/pybikes/aksu.py
+++ b/pybikes/aksu.py
@@ -6,16 +6,15 @@ import re
 import json
 
 from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
-from pybikes.utils import filter_bounds
+from pybikes.utils import Bounded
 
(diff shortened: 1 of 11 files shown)


martgnz:
Nice! this will get rid of the boilerplate. It would be nice if it was automatic but I understand if it isn't possible.

The change looks good, only comment is that you will want to make `Gbfs` a bounded system too.

PR #698, opened 2024-01-03
Add Tashu

martgnz:
A system in Daejeon, South Korea.

It seems to open from March to November: https://bike.tashu.or.kr/main.do?lang=2


pybikes/tashu.py
```diff
@@ -0,0 +1,45 @@
...
+
+class TashuStation(BikeShareStation):
+    def __init__(self, data):
+        super(TashuStation, self).__init__()
+
+        self.name = data["name"]
+        self.latitude = float(data["x_pos"])
+        self.longitude = float(data["y_pos"])
+
+        self.bikes = int(data["parking_count"])
```
martgnz, line 40: there is no dock information :/

PR #700, opened 2024-01-04
Fix Cascavel..?

martgnz:
Cascavel isn't working because when a station is offline the name and id that appear on the page get removed and it just says "offline station". This is not great because names and ids will come empty randomly.. but it let us serve the rest of the system. Thoughts? 

https://www.cascavelvaidebici.mobhis.com.br/

### DIFF
--- a/pybikes/mobhis.py
+++ b/pybikes/mobhis.py
@@ -40,8 +40,14 @@ class MobhisStation(BikeShareStation):
         popup = html.fromstring(info)
         elems = popup.xpath('//text()')
 
+        # in cascavel offline stations lack name and id
         name = elems.pop(0)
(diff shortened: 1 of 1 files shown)


eskerda:
`uid` is not a required field. We ideally want it always to be there, but sometimes it is not possible. In such cases I would go with not getting it at all, and just moving the early return we have in there a bit higher on the code like

```
online = 'Estação offline' not in info
self.extra = {'online': online}

# in cascavel offline stations lack name and id
if not online:
    self.name = None
    self.bikes, self.free = 0, 0
    return

```

martgnz:
This is better indeed. Thanks for checking. I adjusted the logic to do it this way.

eskerda:
Thanks!

PR #701, opened 2024-01-04
Ignore bad points in filter bounds

martgnz:
Bcycle Redding is failing because it was feeding an empty tuple into `filter_bounds` (despite a bbox, not sure what happened).

In any case, this adds a check that ignores empty points before filtering.


eskerda:
Good catch, I think I have found the issue. I used the `or` form, which returns a `False` for `0.0`, so the following ended up returning `(None, None)` instead of `(0.0, 0.0)`.

```
lat = info.get('lat') or status.get('lat')
lng = info.get('lon') or status.get('lon')

>>> print(0.0 or None)
None
```

Let's fix this instead where the bug is

PR #705, opened 2024-01-05
Add Vélozef, Brest

martgnz:
https://www.data.gouv.fr/fr/datasets/velos-a-assistance-electrique-en-libre-service-velozef-sur-brest/#/resources


eskerda:
I see they do not specify any license, so seems good to me. Thanks!

PR #707, opened 2024-01-18
Add Xi'an's system

martgnz:


eskerda:
Hello!

System didn't work for me locally without changing the user agent. We should probably set it to something similar to the client that consumes this feed.

It seems there's a ghost station too in the Mediterranean Sea

martgnz:
Thanks for checking, the new tool is great! I added the bbox.

PR #713, opened 2024-03-04
Youbike: parse bike type

futuretap:
Parsing the number of normal bikes and ebikes.


eskerda:
Looking good, thanks for contributing! 🚀

PR #534, opened 2022-12-27
Add OV Fiets (#309)

eUgEntOptIc44:
fix #309 

Hi @eskerda @ewooonk,

As mentioned in #309 and #499 I'd like to add the Dutch bike sharing provider *OV Fiets* (owned by *Nederlandse Spoorwegen* the national Railways in the Netherlands).

As previously mentioned GBFS for *OV Fiets* are available at https://gbfs.openov.nl/ovfiets. The `stations` feature a `region_id` attribute. Fetching stations (grouped) by 'regions' is unfortunately not possible. Furthermore those regions don't (always) match the (nearest) cities. So manual re-grouping was needed after parsing the data from http://fiets.openov.nl/locaties.json and matching it against the actual stations in https://gbfs.openov.nl/ovfiets/station_status.json.

Regards,

Jean-Luc

Issue #309, opened 2018-02-13
Add OV-fiets (Dutch Railways)


ewooonk:
Bicycle rental service of the Dutch Railways (250+ locations, 12.000+ bikes)

API endpoint: http://fiets.openov.nl/locaties.json

Example JSON Object.

`{"infoImages": [], "name": "UTT - OV-fiets - Utrecht Terwijde", "extra": {"fetchTime": 1518553706, "rentalBikes": "8", "locationCode": "utt001"}, "extraInfo": [], "apps": [], "sites": [], "lat": 52.10017, "link": {"params": {}, "uri": "https://places.ns-mlab.nl/api/v2/places/station-retail/Zelfservice%20OV-fiets%20uitgiftepunt-utt001"}, "stationCode": "UTT", "openingHours": [{"dayOfWeek": 1, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}, {"dayOfWeek": 2, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}, {"dayOfWeek": 3, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}, {"dayOfWeek": 4, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}, {"dayOfWeek": 5, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}, {"dayOfWeek": 6, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}, {"dayOfWeek": 7, "endTime": "00:00", "closesNextDay": true, "startTime": "00:00"}], "lng": 5.04424, "open": "Yes", "thumbnail": {"params": {}, "uri": "api/v1/location/thumbnail/52.10017,5.04424"}, "description": "Utrecht Terwijde"},`

ewooonk:
I propose to use a simple scrapper, e.g. 

https://github.com/eskerda/pybikes/blob/master/pybikes/open_source_bike_share.py

with
```
for station in data:

            longitude = float(station['lng'])
            latitude = float(station['lat'])

            name = station['description']
            free = None
            bikes = int(station['rentalBikes']) 
```

eUgEntOptIc44:
Hi @ewooonk 

You presumably moved on to new topics since your last comment. However I'd like to share a little update on this topic with you.

Since you suggested the integration of *OV Fiets* back in 2018 the provider added GBFS feeds in a public directory at https://gbfs.openov.nl/ovfiets.

[...]

PR #715, opened 2024-03-11
Add OV-fiets

martgnz:
Another attempt at https://github.com/eskerda/pybikes/issues/309.

Rather than dividing the dataset by region, as outlined in https://github.com/eskerda/pybikes/pull/534, this change just adds the system as a whole and relies on the standard GBFS parser.

Let me know what you think!


eskerda:
Closed for https://github.com/eskerda/pybikes/pull/717

PR #720, opened 2024-05-24
Cristolib network is closed

francoisfds:
The "Cristolib" network (France) is definitely closed since 2024-05-20.

See these links (only in french) :
- http://www.cristolib.fr/misc/closed.html (could go offline at some point)
- https://www.ville-creteil.fr/fermeture-definitive-du-service-cristolib

This pull request removes the network from pybikes.

Regards.


eskerda:
Thanks!

PR #727, opened 2024-06-13
Add two Soltráfego bike sharing systems

joaodcp:
Adds Pombike (Pombal) and Bora! (CIM Viseu Dão-Lafões) bikesharing systems. The last one is not yet available but apparently scheduled to be very soon.

There are a bunch other systems using the same provider (Soltráfego) but the endpoint you're using is built for the iframes embedded in each system's website, so the `uid` being used seems to actually be the id of the frame.
Some of them chose not to include that iframe in their websites.

There's another endpoint that allows for the same data to be obtained for every one of the systems supported by this provider. I'm working on those changes, but for now here's these two.

I love this project! 🚀

### DIFF
--- a/pybikes/data/soltrafego.json
+++ b/pybikes/data/soltrafego.json
@@ -29,7 +29,7 @@
                 "city": "CIM Viseu Dão-Lafões",
                 "country": "PT",
                 "latitude": 40.656392,
-                "longitude": -7.912545
+                "longitude": -7.912545,
                 "name": "Bora!"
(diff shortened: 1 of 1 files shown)


eskerda:
Thanks!

> There's another endpoint that allows for the same data to be obtained for every one of the systems supported by this provider. I'm working on those changes, but for now here's these two.

🙏  happy to accept that. Open a PR once you have something going and we can work on merging it

PR #746, opened 2024-09-09
Add Pedalada by Share2Go system

joaodcp:
Hi,

This adds a system deployed in Portugal called Pedalada by a company named Share2Go.

The station address is not included because it takes an extra request per station to retrieve (the details in 'get-station-details-and-ride.php' are not the same as in 'get-station-details.php'). One request per station is already needed to get the number of docked bikes; if you think an extra request per station to get one additional property is worth it then I can implement that.

Thanks again for this great project!

### DIFF
--- a/pybikes/pedalada.py
+++ b/pybikes/pedalada.py
@@ -45,17 +45,17 @@ class PedaladaStation(BikeShareStation):
         super(PedaladaStation, self).__init__()
         self.endpoint = endpoint
 
-        self.name = station['stationName']
-        self.latitude = station['latitude']
-        self.longitude = station['longitude']
+        self.name = data['stationName']
+        self.latitude = data['latitude']
+        self.longitude = data['longitude']
 
         self.extra = {
-            'uid': station['stationNumber'],
-            'slots': station['maximumNumberOfRides'],
-            'online': station['state'] == 1,
-            'photo': station_image_url(endpoint, station['stationNumber']),
-            'open': station['stationIsOpen'] == 1,
-            'in_maintenance': station['stationInMaintenance'] == 1
+            'uid': data['stationNumber'],
+            'slots': data['maximumNumberOfRides'],
+            'online': data['state'] == 1,
+            'photo': station_image_url(endpoint, data['stationNumber']),
+            'open': data['stationIsOpen'] == 1,
+            'in_maintenance': data['stationInMaintenance'] == 1
         }
 
         def update(self, scraper=None):
--- a/pybikes/pedalada.py
+++ b/pybikes/pedalada.py
@@ -60,7 +60,7 @@ class PedaladaStation(BikeShareStation):
 
     def update(self, scraper=None):
         scraper = scraper or PyBikesScraper()
-        station_details_and_rides = scraper.request(station_details_and_rides_url(self.endpoint, self.extra['uid']))
+        station_details_and_rides = json.loads(scraper.request(station_details_and_rides_url(self.endpoint, self.extra['uid'])))
 
         docked_bikes_count = len(station_details_and_rides['rides'])
         self.bikes = docked_bikes_count
--- a/pybikes/pedalada.py
+++ b/pybikes/pedalada.py
@@ -1,0 +1,63 @@
+import json
+
+from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
+
+STATIONS_URL = '{endpoint}/php/get-stations-details-map.php'
+STATION_DETAILS_URL = '{endpoint}/php/get-station-details-and-rides.php?n={station_number}'
+STATION_IMAGE_URL = '{endpoint}/images/stations/station_{station_number}.jpg'
+
+
+def station_details_url(endpoint, station_number):
+    return STATION_DETAILS_URL.format(endpoint=endpoint, station_number=station_number)
+
+def station_image_url(endpoint, station_number):
+    return STATION_DETAILS_URL.format(endpoint=endpoint, station_number=station_number)
+
+class Pedalada(BikeShareSystem):
+    meta = {
+        'company': [
+            'Share2Go - Mobilidade Partilhada, Sociedade Unipessoal Lda'
+        ]
+    }
+
+    def __init__(self, tag, meta, endpoint, bbox=None):
+        meta['company'] += Pedalada.meta['company']
+        super(Pedalada, self).__init__(tag, meta, bounds=bbox)
+        self.endpoint = endpoint
+
+    @property
+    def stations_url(self):
+        return STATIONS_URL.format(endpoint=self.endpoint)
+
+    def update(self, scraper=None):
+        scraper = scraper or PyBikesScraper()
+        stations_data = json.loads(scraper.request(self.stations_url))
+
+        stations = []
+        for station in stations_data:
+            station_details = scraper.request(station_details_url(station['stationNumber']))
+            station = PedaladaStation(station, station_details, self.endpoint)
+            stations.append(station)
+
+        self.stations = stations
+
+
+class PedaladaStation(BikeShareStation):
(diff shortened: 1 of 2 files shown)


eskerda:
Hello! Thanks for contributing.

> The station address is not included because it takes an extra request per station to retrieve (the details in 'get-station-details-and-ride.php' are not the same as in 'get-station-details.php'). One request per station is already needed to get the number of docked bikes; if you think an extra request per station to get one additional property is worth it then I can implement that.

I think it's fine as it is if the extra-extra request only has an address. We have other systems like these on which each station needs a request to get status data. I think it does not need many changes, can you take a look? See: [pybikes/fsm.py](https://github.com/eskerda/pybikes/blob/master/pybikes/fsm.py) and [pybikes/bicincitta.py](https://github.com/eskerda/pybikes/blob/master/pybikes/bicincitta.py)

Let me know if you need anything, ask away

joaodcp:
Hello! Did not notice stations could have an update method too. I'll fix that.

About the address stuff, there are a bunch of "endpoints" for getting data on stations.
Using BIA as an example:
- https://bia.cm-ovar.pt/php/get-station-details.php?n=2 — Station details with address but no details on how many bikes are docked.
- https://bia.cm-ovar.pt/php/get-station-rides.php?n=2 — Get the bikes on a station
- https://bia.cm-ovar.pt/php/get-station-details-and-rides.php?n=2 — get a few details (which do not include the address) and the docked bikes.

I'm only using the last one because it includes most info, including bikes.

Also there's some stations with a schedule, I don't know how you'd like me to format that. I was thinking of formatting it like the opening_hours property on OSM (00:00 - 23:59). Let me know what you think I should do. Thanks!

pybikes/pedalada.py
```diff
@@ -0,0 +1,72 @@
...
+            station_details_and_rides = scraper.request(station_details_and_rides_url(self.extra['uid']))
+
+            docked_bikes_count = len(station_details_and_rides['rides'])
+            self.bikes = docked_bikes_count
+            self.free = self.extra['slots'] - docked_bikes_count
+
+            station_details = station_details_and_rides['details'][0]
+            self.extra['online'] = station_details['state'] == 1
+            self.extra['open'] = station_details['stationIsOpen'] == 1
+            self.extra['in_maintenance'] = station_details['stationInMaintenance'] == 1
```
eskerda, line 72: Seems this part needs to be indented back 4 spaces to work. As it is, it's declared inside `__init__` and has no effect

pybikes/pedalada.py
```diff
@@ -0,0 +1,72 @@
...
+            'slots': data['maximumNumberOfRides'],
+            'online': data['state'] == 1,
+            'photo': station_image_url(endpoint, data['stationNumber']),
+            'open': data['stationIsOpen'] == 1,
+            'in_maintenance': data['stationInMaintenance'] == 1
+        }
+
+        def update(self, scraper=None):
+            scraper = scraper or PyBikesScraper()
+            station_details_and_rides = scraper.request(station_details_and_rides_url(self.extra['uid']))
```
eskerda, line 63:
```suggestion
            station_details_and_rides = scraper.request(station_details_and_rides_url(self.endpoint, self.extra['uid']))
```

eskerda (changes requested):
Couple of issues with the implementation. We can keep reviewing after these are fixed. Thanks!

You can try if this implementation passes the update tests by running the following.

```
pytest -k pedalada -vvv
```

More info on running tests [here](https://github.com/eskerda/pybikes/blob/master/README.md#tests)

joaodcp:
Yup, my bad. All fixed now! Thanks again

joaodcp:
Yup, my bad. All fixed now! Thanks again

pybikes/pedalada.py
```diff
@@ -0,0 +1,72 @@
...
+            'slots': data['maximumNumberOfRides'],
+            'online': data['state'] == 1,
+            'photo': station_image_url(endpoint, data['stationNumber']),
+            'open': data['stationIsOpen'] == 1,
+            'in_maintenance': data['stationInMaintenance'] == 1
+        }
+
+    def update(self, scraper=None):
+        scraper = scraper or PyBikesScraper()
+        station_details_and_rides = scraper.request(station_details_and_rides_url(self.endpoint, self.extra['uid']))
```
eskerda, line 63: string needs to be parsed as JSON

pybikes/pedalada.py
```diff
@@ -0,0 +1,72 @@
...
+        self.latitude = data['latitude']
+        self.longitude = data['longitude']
+
+        self.extra = {
+            'uid': data['stationNumber'],
+            'slots': data['maximumNumberOfRides'],
+            'online': data['state'] == 1,
+            'photo': station_image_url(endpoint, data['stationNumber']),
+            'open': data['stationIsOpen'] == 1,
+            'in_maintenance': data['stationInMaintenance'] == 1
```
eskerda, line 58: Do you know the difference between `state`, `stationIsOpen` and `stationInMaintenance`?

In vitelinhas there's one station with the following:

```
{
  "name": "Avenida da Liberdade",
  "latitude": 41.45442,
  "longitude": -8.17702,
  "bikes": 3,
  "free": 8,
  "extra": {
    "uid": 2,
    "slots": 11,
    "online": false,
    "photo": "https://vitelinhas.cm-fafe.pt/images/stations/station_2.jpg",
    "open": true,
    "in_maintenance": false
  },
  "type": "station",
  "tag": "vitelinhas"
}
```

On the website that's reported as inactive. So my take on there is that open maybe means it's within the schedule opening hours?

eskerda: My take is that it's a bit confusing and we should try to find a better name for it, or decide to not include it if we skip schedules altogether

joaodcp: > Do you know the difference between `state`, `stationIsOpen` and `stationInMaintenance`?
> 
> In vitelinhas there's one station with the following:
> 
> ```
> {
>   "name": "Avenida da Liberdade",
>   "latitude": 41.45442,
>   "longitude": -8.17702,
>   "bikes": 3,
>   "free": 8,
>   "extra": {
>     "uid": 2,
>     "slots": 11,
>     "online": false,
>     "photo": "https://vitelinhas.cm-fafe.pt/images/stations/station_2.jpg",
>     "open": true,
>     "in_maintenance": false
>   },
>   "type": "station",
>   "tag": "vitelinhas"
> }
> ```
> 
> On the website that's reported as inactive. So my take on there is that open maybe means it's within the schedule opening hours?

open means that the station would be open according to its schedule. But it can be unavailable due to connection problems but would be open per se. On the websites, the station being offline takes precedence over the schedules and if it is offline it's just shown as inactive. I believe a station in maintenance means that it is not available at the moment, due to being serviced. Maybe we could combine these all into one property.

eskerda: It's partly fault on pybikes side that we do not have a full specification for such cases. For some we have a `status` field that is a kind of all goes in. We could try to merge `open` and `in_maintenance` with the `status` field with one of the following `open`, `closed`, `maintenance`. Since we can consider that a station that is in maintenance is not open nor closed per se.

```
def format_status(self, data):
    if data['stationInMaintenance']:
        return 'maintenance'
    elif data['stationIsOpen']:
        return 'open'
    else:
        return 'closed'

[...]

'status': self.format_status(data),
```

Alternatively we can go with a oneliner but it's a bit of weird python

```
'status': 'maintenance' if data['stationInMaintenance'] else 'open' if data['stationIsOpen'] else 'closed'
```

eskerda:
@joaodcp any progress on this PR? I feel we are good to merge once this https://github.com/eskerda/pybikes/pull/746#discussion_r1759720075 goes in

eskerda:
Integrated changes from this branch over https://github.com/eskerda/pybikes/pull/755

PR #750, opened 2024-09-12
do not parse cookies on DB requests

### LINKED ISSUE #749: Outdated station information for Call a Bike (Germany)
I have checked daily during the past days, and the information on citybik.es is always wrong compared to the Call a Bike App or informaiton on Google Maps. Stations that contain bikes are shown as empty, and those that are empty show as having bikes available.

To see yourself, simply zoom in on any city (for example Munich), search for call a bike and compare both maps.

Example: Station S+U Karlsplatz / Bayerstraße, 12.09.24  - 6 bikes according to citybik.es and 9 bikes according to the Call A Bike App/Google Maps:

image

image

eskerda:
requests with Cookie headers set after Set-Cookie responses where being
rejected by DB gateway

Fix #749

### DIFF
--- a/pybikes/deutschebahn.py
+++ b/pybikes/deutschebahn.py
@@ -38,6 +38,7 @@ class DB(Gbfs):
 
     def update(self, scraper=None):
         scraper = scraper or PyBikesScraper()
+        scraper.parse_cookies = False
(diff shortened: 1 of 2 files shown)


eskerda:
image

PR #753, opened 2024-09-17
check changes

eskerda:
Adds a workflow that runs instance tests on instances that have changed on this particular branch, compared to `origin/master`.

As it is now, it runs as a separate workflow so it gets its own report summary. It could potentially be included on the regular workflow, but that would collide with the full instance report.

Demo run: https://github.com/eskerda/pybikes/commit/1c1942ac02c94bab456ef1fe03f06ec5bd90b8e3


eskerda:
Tried with a commit https://github.com/eskerda/pybikes/pull/753/commits/2ba5b287f99523dab5ab6e5b632640e0ab67fbad that changed several networks and got this report: https://github.com/eskerda/pybikes/actions/runs/10904691066/job/30261737238?pr=753

image

So this workflow seems to work ok :)

PR #769, opened 2024-10-15
Add meioB

### LINKED ISSUE #742: Support meioB, Portugal
https://meiob.pt/mapa-rede/
https://play.google.com/store/apps/details?id=com.cme.meiob&hl=de&pli=1

martgnz:
Multi-city system in the region of Médio Tejo, close to Lisbon.

Closes #742.

Url: https://meiob.pt/mapa-rede/

### DIFF
--- a/pybikes/meiob.py
+++ b/pybikes/meiob.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2024, Martín González Gómez <m@martingonzalez.net>
+# Distributed under the AGPL license, see LICENSE.txt
+
+import re
+
+from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
+
+
+class Meiob(BikeShareSystem):
+
+    def __init__(self, tag, meta, endpoint):
+        super(Meiob, self).__init__(tag, meta)
+
+        self.endpoint = endpoint
+
+    def update(self, scraper=None):
+        scraper = scraper or PyBikesScraper()
+
+        stations = []
+        raw = scraper.request(self.endpoint)
+        data = zip(
+            re.findall(r'stationlat\">(.*?)</div>', raw, re.DOTALL),
+            re.findall(r'stationlon\">(.*?)</div>', raw, re.DOTALL),
+            re.findall(r'pe-xxl-3\">\n\s+<h2>(.*?)</h2>', raw, re.DOTALL),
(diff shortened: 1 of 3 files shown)


pybikes/meiob.py
```diff
@@ -0,0 +1,64 @@
...
+        raw = scraper.request(self.endpoint)
+        data = zip(
+            re.findall(r'stationlat\">(.*?)</div>', raw, re.DOTALL),
+            re.findall(r'stationlon\">(.*?)</div>', raw, re.DOTALL),
+            re.findall(r'pe-xxl-3\">\n\s+<h2>(.*?)</h2>', raw, re.DOTALL),
+            re.findall(r'Bicicletas disponíveis</p>\n\s+<p class=\"mb-0 bike__available\">(.*?)</p>', raw, re.DOTALL),
+            re.findall(r'Docas disponíveis</p>\n\s+<p class=\"mb-0 bike__available\">(.*?)</p>', raw, re.DOTALL),
+            re.findall(r'<div class=\"w-100\">\n\s+<p class=\"mb-0 bike__distance__title text-center\">(.*?)</p>', raw, re.DOTALL),
+            re.findall(r'</p>\n\s+<p class=\"mb-0 bike__distance__title text-center\">(.*?)</p>', raw, re.DOTALL)
+        )
```
eskerda, line 30: We really should be using xpath selector for these, smth like

```python
        html = lxml.html.fromstring(raw)
        stations = html.xpath("""
            //div[contains(@class, "bikes")]
                /div[contains(@class, "bike")]
        """)

        for station_element in stations:
            station = MeiobStation(station_element)
```

And then on the element parsing itself

```python
class MeiobStation(BikeShareStation):
    def __init__(self, element):
        name = element.xpath('.//h2//text()')[0]
        lat = element.xpath("""
            ./div[contains(@class, 'stationlat')]//text()
        """)[0]
        ...
```

If you prefer to use css selectors, we can add `cssselect` to the project so we can do

```python
stations = html.cssselect('.bikes > .bike')
...
lat = element.cssselect('div.stationlat')[0].text
```

martgnz: would that be ok? xpath is very hard to use when it's not straightforward ids. css would be way easier to understand and to use.

eskerda: lxml has cssselect support through the cssselect library that has no extra dependencies, so yeah, that's ok. Just add it to `setup.py`.

martgnz: done, let me know what you think!

eskerda: looking good!

Issue #775, opened 2024-12-07
New nextbike services on spain


Gabriel44200:
TorrentBici (Torrent): domain en
Bicilog (Logroño):domain ej

PR #776, opened 2024-12-07
Move torrentbici to nextbike

martgnz:
Torrentbici has "graduated" from being part of mibisivalencia to a Nextbike service. They removed it from the map and now they have their own section of the site.

This change implements that and moves the coords of mibisivalencia since they were centered around Torrent before.

https://torrentaldia.com/torrentbici-se-renueva-con-100-bicis-electricas-y-19-estaciones/
https://www.mibisivalencia.es/torrentbici/mapa.php
https://www.mibisivalencia.es/mapa/mapa.php

Part of https://github.com/eskerda/pybikes/issues/775.

### DIFF
--- a/pybikes/data/movus.json
+++ b/pybikes/data/movus.json
@@ -8,8 +8,8 @@
                 "city": "Valencia, área metropolitana",
                 "country": "ES",
                 "name": "MIBISI",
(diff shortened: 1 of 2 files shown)


eskerda:
Thanks!

PR #777, opened 2024-12-07
Move bicilog to Nextbike

martgnz:
Bicilog has also moved, this time from Bicicard to Nextbike.

https://www.nextbike-bicilog.com/es/
https://logrono.es/-/el-servicio-de-bicilog-se-amplia-con-200-nuevas-bicicletas-50-de-ellas-electricas-y-ya-puede-ser-utilizado-por-toda-la-ciudadania

Part of #775.


eskerda:
Thanks. Let's 🙏  all itcl systems end up in nextbike

PR #781, opened 2025-01-13
Change Caen provider

Ephtolens:
Update the BSS provider for caen (cyckleo -> fifteen)


Ephtolens:
GBFS from Aachen (DE) and Le Locle (CH) seems broken and cause a test to fail

eskerda:
> GBFS from Aachen (DE) and Le Locle (CH) seems broken and cause a test to fail

Yeah, that's fine. Thanks for fixing Caen!

PR #786, opened 2025-01-16
Retire Youbike v1 feed

martgnz:
This system used two feeds, the "v1" version and the "v2" versions. They were serving the same data format but v1 was generated by their first generation of bikes/app in some cities and v2 serves the newest version, which was more widespread if i recall correctly. Unfortunately the whole build was failing because the v1 feed stopped working.

Looking at their sites (https://www.youbike.com.tw/region/main/) all the pages are now calling the v2 feed (e.g. https://www.youbike.com.tw/region/taipei/stations/list/) so I think it's safe to remove v1 from the call. My guess is that they managed to move all systems to this feed.

### DIFF
--- a/pybikes/youbike.py
+++ b/pybikes/youbike.py
@@ -8,7 +8,6 @@ from pybikes import BikeShareSystem, BikeShareStation, PyBikesScraper
 
 
 class YouBikeTw(BikeShareSystem):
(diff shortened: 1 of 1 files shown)


eskerda:
Seems they retired it on the 31th of dec. Data looks the same, so I think they probably migrated all to v2. Thanks for checking this!

PR #789, opened 2025-04-01
fix(vcub): Switch to gbfs

Ephtolens:
- Switch Bordeaux VLS from Keolis to GBFS (still keolis operator)
- Added support for GBFS v3 since num_bikes_available is not present anymore in station_status.json
- Remove gbfs that are empty or did not work anymore (502 bad gateway)

### DIFF
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -1160,54 +1160,6 @@
             "feed_url": "https://santana.publicbikesystem.net/customer/gbfs/v2/gbfs.json"
         },
         {
-            "tag": "nicosia",
-            "meta": {
-                "latitude": 35.1886,
-                "longitude": 33.3570,
-                "city": "Nicosia",
-                "name": "Nicosia Bike Share",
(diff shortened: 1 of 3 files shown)


Ephtolens:
You're right, its a fix not a feat, i wrote too fast !

eskerda:
> You're right, its a fix not a feat, i wrote too fast !

I meant to change it on the commit messages themselves. Just rebase and remove the prefix if it's not relevant to the changes (ie: vcub has no relation to removing dead gbfs, adding partial support for gbfs is arguably a feat on gbfs, etc).

All these commit prefixes can get so many interpretations that my decision is to just **get rid of them**. I do not want to waste time discussing if something is a `fix` or a `feat` or a `chore`. If you want to include the implementation name in front of the commit, like: `[gbfs] add partial support for v3` that's fine but **not mandatory**.

Ephtolens:
Here is the new PR : https://github.com/eskerda/pybikes/pull/790

PR #790, opened 2025-04-03
[gbfs] Remove old gbfs, change vcub source and add support for gbfs v3

Ephtolens:
Here is the new PR, sorry for the inconvenience !

### DIFF
--- a/pybikes/gbfs.py
+++ b/pybikes/gbfs.py
@@ -22,7 +22,7 @@ def get_text(text):
         return text
     if isinstance(text, list):
         return next(iter(text))['text']
-    return ""
+    return None
 
 
 class Gbfs(BikeShareSystem):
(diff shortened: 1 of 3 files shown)


pybikes/gbfs.py
```diff
@@ -17,6 +17,14 @@
     unicode = str
 
 
+def get_text(text):
+    if isinstance(text, str):
+        return text
+    if isinstance(text, list):
+        return next(iter(text))['text']
+    return ""
```
eskerda, line 25: I prefer this function to return `None` and cascade a failure than hiding the error with an empty string.

```suggestion
    return None
```

pybikes/gbfs.py
```diff
@@ -242,8 +250,11 @@ def __init__(self, info, vehicles_info):
         if not info['is_installed']:
             raise exceptions.StationPlannedException()
 
-        self.name = unicode(info['name'])
-        self.bikes = int(info['num_bikes_available'])
+        self.name = unicode(get_text(['name']))
```
eskerda, line 253:
```suggestion
        self.name = unicode(get_text(info['name']))
```

PR #794, opened 2025-04-06
Shutdown MuyBici

martgnz:
MuyBici in Murcia has been shut down. This system only had that service. There's a new temporary contract with Dott but I'm not sure yet if they're docked or dockless.

https://www.laopiniondemurcia.es/murcia/2025/01/16/muybici-echa-persiana-febrero-llegaran-113422058.html
https://www.orm.es/noticias-2024/comienza-el-nuevo-servicio-de-alquiler-de-bicicletas-de-murcia/
https://www.interempresas.net/Smart_Cities/Articulos/584554-El-Ayuntamiento-de-Murcia-pone-en-marcha-un-proyecto-piloto-con-700-bicicletas-electricas.html


eskerda:
We can later add dott systems once we land dockless support. Going to add an issue so we don't forget Murcia :)

Issue #791, opened 2025-04-03
handle GBFS versions


eskerda:
Instead of going all in try to handle different GBFS versions in separate implementations.

PR #797, opened 2025-04-07
Update Ecovelo feed

martgnz:
They moved to Gbfs v3 and the 2.2 feed was dropped, breaking these twenty three systems.

We can hardcode v3 in the url too but I fear that at some point this will be upgraded and will break again. The url without the version seems to redirect to the latest feed.

### DIFF
--- a/pybikes/ecovelo.py
+++ b/pybikes/ecovelo.py
@@ -11,7 +11,7 @@ class Ecovelo(Gbfs):
         'company': ['Ecovelo'],
     }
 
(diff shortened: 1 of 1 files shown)


eskerda:
Yup.. it's weird, because the root `gbfs.json` links to non working versions. Generally I would prefer to pin it to at least some version, but I agree with you.

eskerda:
Once we work on https://github.com/eskerda/pybikes/issues/791 and handle versions based on the response this won't be a problem (we can explicitly fail with unhandled versions).

PR #798, opened 2025-04-07
Update Cyclopolis

martgnz:
Lots of systems were turned off from the feeds on June 2024. They might have migrated somewhere else but they all redirect to the main site on https://www.cyclopolis.gr/index.php/el/cyclopolis-per-municipality.

I discovered a new endpoint for Chania and made a new class, which is also used by the system in the Agricultural University of Athens.

### DIFF
--- a/pybikes/cyclopolis.py
+++ b/pybikes/cyclopolis.py
@@ -9,8 +9,6 @@ import json
 from pybikes import BikeShareSystem, BikeShareStation
 from pybikes.utils import PyBikesScraper
 
-__all__ = ['Cyclopolis', 'CyclopolisStation']
-
 LAT_LNG_RGX_GOOGLE = r'latLng:\[(\d+.\d+).*?(\d+.\d+)\]'
 LAT_LNG_RGX_MAPBOX = r'"lat":\s?"(\d+.\d+).*?"lon":\s?"(\d+.\d+)'
 DATA_RGX = r'data"?\:.*?<span.*?>(.*?)<\\?/span>'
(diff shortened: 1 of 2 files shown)


pybikes/cyclopolis.py
```diff
@@ -98,3 +99,46 @@ def __init__(self, name, latitude, longitude, bikes, free, extra):
...
+        data = json.loads(scraper.request(self.feed_url))
+
+        stations = []
+
+        self.stations = []
+        for station in data:
+            name = station["name"]
+            latitude = float(station["location"]["coordinates"][1])
+            longitude = float(station["location"]["coordinates"][0])
+            bikes = station["vehicle_docks_available"][0]["count"]
```
martgnz, line 123: this is ambiguous and might be only slots. i haven't seen more in the feed.

eskerda: Yes.. this seems like a direct port of a gbfs v3 feed, and `vehicle_docks_available` counts the number of docks per vehicle type. The number of bikes available should be under `num_vehicles_available` or `vehicle_types_available` which is empty, meaning one of the following:
* it is really empty
* they do not provide this information

I think we should set this as the number of `free` on the station, and maybe wait and see if `vehicle_types_available` comes alive at some point?

eskerda: Added handling for vehicle types and vehicle docks. 

```diff
-            bikes = station["vehicle_docks_available"][0]["count"]
+            bikes = sum(map(lambda vd: vd['count'], station["vehicle_types_available"]))
+            free = sum(map(lambda vd: vd['count'], station["vehicle_docks_available"]))
```

Will add an issue to check back in some weeks https://github.com/eskerda/pybikes/issues/801

PR #802, opened 2025-04-07
Add name to Cyclopolis system

martgnz:
It's for the Agricultural University of Athens, not just "Athens".


eskerda:
Does it make sense to set it up as the name of the system and not the city? I know we are using the `city` field for many things, like a group of cities or regions, but this sets up a precedent that the city field really does not mean anything.

martgnz:
Sorry these systems had no name so I forgot about that field. Updated.

PR #814, opened 2025-05-02
Correct meinRad coordinates

### LINKED ISSUE #813: meinRad from Mainz does not appear on the Map
When trying to check for Meinrad in Mainz (Germany) I wasn't able to find it on the map, although it is available in the configuration. So there might be some issue. Would also be happy to help fixing it, if there is an easy way to check what's wrong with it. Its based on Nextbike so a GBFS feed is available at https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_mz/gbfs.json

1Maxnet1:
Choose the coordinates of the company providing the bikesharing service

Resolves #813

### DIFF
--- a/pybikes/data/nextbike.json
+++ b/pybikes/data/nextbike.json
@@ -2598,8 +2598,8 @@
                 "name": "meinRad MVG Mainz",
                 "city": "Mainz",
                 "country": "DE",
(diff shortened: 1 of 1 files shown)


eskerda:
Perfect. Thank you @1Maxnet1!

PR #822, opened 2025-07-08
[nextbike] fix wrm-wroclaw uid

eskerda:


eskerda:
Fixed existing systems in Poland and also added some new ones.

Lublin has moved to another system

Torvelo, assumedly, is hibernating ?

eskerda:
There are still failing systems in nextbike and gbfs, but let's keep this PR just about addressing feeds in PL.

PR #825, opened 2025-07-08
Fix failing GBFS feeds

eskerda:
Fifteen updated feed urls so the ones we had pointing to their resolved url failed.


eskerda:
All fixable are fixed. The failing one in Columbus seems to be a flaky system reporting stations in `station_status` but not in `station_information`

`¯\_(ツ)_/¯`

PR #830, opened 2025-08-02
Opensource Bikesharing v2

sveneld:
Bike sharing system WhiteBikes  in Bratislava migrated to v2 version.
Information about stands and bikes is now provided by token


sveneld:
https://github.com/cyklokoalicia/OpenSourceBikeShare/blob/main/Access%20the%20data.md will be added soon
You can currently write to email

PR #844, opened 2025-09-16
Adding Graoulib (Metz, France) using Fifteen

rogafe:
Hi, I added GraouLib’ for Metz


eskerda (changes requested):
Hello @rogafe. Let's either add the gbfs feed or the fifteen feed, but not both!

PR #863, opened 2025-11-18
Add Graoulib GBFS (Metz, FR)

Ephtolens:

PR #864, opened 2025-11-25
integrate changes from  #863

eskerda:
Integrate changes from https://github.com/eskerda/pybikes/pull/863

@Ephtolens your master was outdated so the merge was messy. I have integrated your changes plus added a couple of fixes i intended to review.

For the feed on montpellier, we can hardcode the urls for now, so the parser does not try to reach for free_bikes feed

Thanks!

### DIFF
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -1212,18 +1212,6 @@
             "feed_url": "https://fortworth.publicbikesystem.net/customer/gbfs/v2/gbfs.json"
         },
         {
-            "tag": "pony-nice",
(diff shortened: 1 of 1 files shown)


Ephtolens:
Actually, after asking Pont,  we've obtained an apikey at Geovelo but we're not alloweed to share it publicly for now, i'll ask them if i can add it to this repo.
I've been thinking about a new type of feed : authed_gbfs that overload the gbfs class and use the authed param and overload the get feeds fonction to add the apikey provided to the gbfs url
What do you think of it ?

PR #871, opened 2026-01-27
feat(cyclocity): Switch to GBFS and update some outdated tags

Ephtolens:
JCDecaux (Operator of the Cyclocity BSS brand) recently released their GBFS feed instead of their proprietary feed.
Here's the switch to GBFS feed ! 
https://developer.jcdecaux.com/#/gbfs?page=usage
I'im wondering if renaming the tag is a good idea (for Besançon and Nantes). It's more accurate but it could brake things if people upgrade their Pybike version without having a look

### DIFF
--- a/pybikes/data/cyclocity.json
+++ b/pybikes/data/cyclocity.json
@@ -1,215 +1,218 @@
 {
     "system": "cyclocity",
-    "class": "Cyclocity",
-    "instances": [
-        {   
-            "contract": "amiens",
-            "tag": "velam",
-            "meta": {
-                "city": "Amiens",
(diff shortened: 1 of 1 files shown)


eskerda:
> Hmm, i've made the change but the error for "Instance tests (only changes) is weird

I have removed python 3.9 from the workflow here https://github.com/eskerda/pybikes/pull/872

Can you rebase against current `master` and force push to see if it makes the problem go away?

Thanks!

Ephtolens:
Great catch ! I'll reach them out, let's stand by on this PR fornow

PR #873, opened 2026-01-28
cyclocity gbfs

eskerda:
Cyclocity has just announced a new public GBFS endpoint (and decommission of the old feed on 12/31/2026). This PR moves cyclocity instances to use the new GBFS feed.

It also hardens parsing of `last_reported` field in GBFS, since some stations in a cyclocity feed included huge negative ints to signal an unknown status. These huge ints were breaking serialization when running tests in parallel (see https://github.com/eskerda/pybikes/pull/871). Accepting arbitrary input for last_reported was a bad idea anyway.

### DIFF
--- a/pybikes/gbfs.py
+++ b/pybikes/gbfs.py
@@ -8,6 +8,7 @@
 import json
 from warnings import warn
 from urllib.parse import urljoin, urlparse, parse_qs
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


Ephtolens:
My bad ! I never saw that the contract field was always there all this time

PR #876, opened 2026-02-04
feat(pau): Move Pau from cyckleo to fifteen

Ephtolens:
New feed for Pau (FR)

### DIFF
--- a/pybikes/data/cykleo.json
+++ b/pybikes/data/cykleo.json
@@ -12,17 +12,6 @@
             }
         },
         {
-            "organization": 3,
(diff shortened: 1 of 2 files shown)


eskerda:
> It looks like the form factor "other" in vehicle_types is not handle, like in this feed : https://stables.donkey.bike/api/public/gbfs/2/donkey_neuchatel/en/vehicle_types.json

We might not, but it is a warning not an error. Does this happen with this feed in Pau or is this about other feeds?

eskerda:
Thanks @Ephtolens!

PR #898, opened 2026-08-25
[vlille] Move to GBFS endpoint

### LINKED ISSUE #897: Outdated data returned again by the API for "vlille" network
Hello,

It appears that the data for the "vlille" network as reported by the API is outdated -- see https://api.citybik.es/v2/networks/vlille where for instance station 0 has timestamp "2026-05-11T06:24:19.678388+00:00Z".

A similar problem occurred in the past (and had been fixed then): #779 .

Would it be possible to fix this issue again? Many thanks if that's possible, and I hope the report is useful!

martgnz:
Looks like the moved to a GBFS endpoint and abandoned the other one.

Found here: https://transport.data.gouv.fr/datasets/ilevia-vlille-disponibilite-en-temps-reel

Closes https://github.com/eskerda/pybikes/issues/897.

### DIFF
--- a/pybikes/data/gbfs.json
+++ b/pybikes/data/gbfs.json
@@ -1558,7 +1558,12 @@
                 "name": "V'Lille",
                 "longitude": 3.057256,
                 "company": ["Keolis", "ilévia"],
-                "country": "FR"
+                "country": "FR",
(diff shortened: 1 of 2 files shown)


eskerda:
@martgnz any license available? we can add a source and license like in https://github.com/eskerda/pybikes/blob/3d1027c27d066b0bb854ff062d17bf353e7025cc/pybikes/data/gbfs.json#L511-L515

